PR #5, opened 2022-02-02
#514 Refactor GlspElkLayoutEngine

tortmayr:
- Refactor the `GlspElkLayoutEngine` and related concepts to properly support layout computation for ports. In this process we also opted out of trying to reuse the existing `SprottyElkLayoutEngine` and instead implement a new layout engine specifically for the GLSP graphmodel. This implementation is more in line with the Java ELK-Layout Engine and implements similar features like automatic detection of common ancestors of edge source/targets to correctly transform them in the ELK graph.

- Removes the no longer needed `BasicTypeMapper`

Also: 
- Provide utility functions for querying parent elements of a given element that match a predicate. (gmodel-util.ts)
- Only rerexport the the subset of types that are really needed from "sprotty-elk" to avoid unncessary naming clashes.
- Add jsdoc for `GModelSerializer`
- Fix implementation of `getAllEdges` in `GModelIndex`
- Remove unused (and unset) `source` and `target` references in `GEdge`


packages/layout-elk/src/glsp-elk-layout-engine.ts
```diff
@@ -13,45 +12,321 @@
...
+
+    protected transformGraph(graph: GGraph): ElkGraphElement {
+        const elkGraph: ElkNode = {
+            id: graph.id,
+            layoutOptions: this.configurator.apply(graph)
+        };
+        if (graph.children) {
+            elkGraph.children = this.findChildren(graph, GNode).map(child => this.transformToElk(child)) as ElkNode[];
+            elkGraph.edges = [];
+            this.elkEdges.push(...(this.findChildren(graph, GEdge).map(child => this.transformToElk(child)) as ElkEdge[]));
```
sgraband, line 148: Should these edges not also be pushed into the `elkGraph.edges`? Otherwise the array is always empty, right?

packages/layout-elk/src/glsp-elk-layout-engine.ts
```diff
@@ -13,45 +12,321 @@
...
+    protected transformNode(node: GNode): ElkNode {
+        const elkNode: ElkNode = {
+            id: node.id,
+            layoutOptions: this.configurator.apply(node)
+        };
+
+        if (node.children) {
+            elkNode.children = this.findChildren(node, GNode).map(child => this.transformToElk(child)) as ElkNode[];
+            elkNode.edges = [];
+            this.elkEdges.push(...(this.findChildren(node, GEdge).map(child => this.transformToElk(child)) as ElkEdge[]));
```
sgraband, line 164: Same here

sgraband (commented):
Thanks @tortmayr!

Changes look good. Just a few inline comments below.

packages/layout-elk/src/element-filter.ts
```diff
@@ -0,0 +1,91 @@
...
+
+    protected filterEdge(edge: GEdge): boolean {
+        const source = this.modelState.index.get(edge.sourceId);
+        if (!source) {
+            return false;
+        }
+
+        if ((source instanceof GNode && !this.filterNode(source)) || (source instanceof GPort && !this.filterPort(source))) {
+            return false;
+        }
```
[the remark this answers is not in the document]
tortmayr: I agree. We can even omit the undefined check entirely because the following instanceof check will be false anyways if the source is not defined.

packages/layout-elk/src/glsp-elk-layout-engine.ts
tortmayr: No, edges need special handling. ELK expects that edges are always set in their respective parent container (either the graph or another node in case of nested diagrams). Otherwise edges want be layouted correctly. However, the same assumption is not true for the graphical model. If we have an edge between A and B the edge doesn't have to be a child of the container of A and B. It could be placed anywhere in the graph (e.g. as child of A).

Hence I implemented the same mechanism that is used in the Java LayoutEngine. (Store the edges in a temporary map and then in a postprocess step find the most suitable parent container). (https://github.com/eclipse-glsp/glsp-server-node/blob/54b5138f7867d0a73862e5bbac0d350deca8ec3f/packages/layout-elk/src/glsp-elk-layout-engine.ts#L76)

sgraband (approved):
LGTM :+1:

PR #10, opened 2022-03-10
Remove yarn.lock from gitignore

tortmayr:
Fixes eclipse-glsp/glsp/issues/581


planger (approved):
Thanks, looks good to me! Build works now :+1:

PR #11, opened 2022-03-25
Refactorings: Source model loading and saving

planger:
- Rename ModelSourceLoader to SourceModelPersistence
  - Add saving to the SourceModelPersistence interface
  - Delegate from the SaveModelActionHandler to that implementation
  - Update javadoc and variable names where fit

https://github.com/eclipse-glsp/glsp/issues/582


packages/server-node/src/base-impl/gmodel-storage.ts
```diff
@@ -72,4 +74,14 @@ export class GModelLoader implements ModelSourceLoader {
             throw new GLSPServerError(`Could not read & parse file contents of '${url}' as json`, error);
         }
     }
+
+    saveSourceModel(action: SaveModelAction): MaybePromise<void> {
+        try {
+            const data = this.modelSerializer.createSchema(this.modelState.root);
+            // eslint-disable-next-line no-null/no-null
+            writeFileSync(this.modelState.sourceUri!, JSON.stringify(data, null, 2));
```
tortmayr, line 82: Instead of disabling the eslint rule you can just use `undefined` here instead of null
```suggestion
            writeFileSync(this.modelState.sourceUri!, JSON.stringify(data, undefined, 2));
```

tortmayr (approved):
Thanks looks good to me.

packages/server-node/src/base-impl/gmodel-storage.ts
planger: Ah right, yeah I just moved the code. I amended your change. Thanks!

PR #13, opened 2022-04-15
Explicit methods for mandatory bindings

planger:
This change sketches an alternative approach to explicitly force mandatory bindings in diagram modules.

https://github.com/eclipse-glsp/glsp/issues/620

### DIFF
--- a/examples/workflow-server/src/workflow-diagram-module.ts
+++ b/examples/workflow-server/src/workflow-diagram-module.ts
@@ -17,7 +17,6 @@ import {
     ClassMultiBinding,
     CommandPaletteActionProvider,
     ContextMenuItemProvider,
-    DefaultToolPaletteItemProvider,
     DiagramConfiguration,
     GLSPServer,
     GModelDiagramModule,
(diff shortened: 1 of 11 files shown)


tortmayr:
@planger  I have pushed a new proposal using `BindingTargets` (as discussed in  in https://github.com/eclipse-glsp/glsp/issues/620)
WDYT?

planger:
Oh, nice! Yeah that looks great actually! Thanks a lot!

tortmayr:
Thanks for the feedback. So should we go ahead and introduce explicit method bindings for all diagram module bindings?
Then the Java and Node API would be more or less aligned.
Adopters can then  chose whether they want to rebind certain services by overwriting  the  corresponding method or if they wan to go the `standard` inversify route and implement and load a new module that provides the rebinds.

planger:
Yes, I'd be in favor of that. But of course, as always it is a matter of time. If it fits nicely with preparing the getting-started examples, that'd be great. But I'd consider those to have higher prio.

Thanks!

PR #15, opened 2022-04-22
#620 Introduce explicit binding methods for diagram module

tortmayr:
Introduce the concept of `BindingTargets` which enables us to define bindings in the diagram and server modules in a more explicit way using dedicated submethods. (Similar to the pattern we currently use in the Java  GLSP Server).  Using `BindingTargets` allows us to move the specification of the target binding into a
submethod without limiting the power of inversify. Adopters can still chose whether they want to bind a service identifier to constructor, self, constant value or another service.

In addition, this change fixes a couple of issues in the test setup regarding error handling.

Fixes eclipse-glsp/glsp/issues/629


planger (approved):
Thanks, looks great! I think it is much clearer and more explicit now what you are supposed to implement! :+1:
Tested the workflow example and it seems to work as expected.

PR #16, opened 2022-04-25
General refactoring and cleanup

tortmayr:
- Ensure consistent usage of `ModelState`
Previously the default implementations used and inconsistent mix of injecting `GModelState` (the custom model state for the direct Gmodel lib)
and `ModelState` (the default class). This introduced and indirect dependency to the `GModelDiagramModule`.  Since the `GModelState` actually did only contain generally reusable code it has been removed and merged with `ModelState`. All default implementations now consistently inject the `ModelState`.

- Move all services (action & operations handlers etc.) that contain implementation specific code for the direct gmodel usecase (graphical model === source model) into a dedicated `gmodel-lib` subdirectory.

- Extract a reusable `AbstractJsonModelStorage` base implementation from the `GModelStorage` that can be reused for any model storage
that loads JSON models from files.

- Introduce a generic abstract base implementation for `CreateNodeOperationHandlers`. 

- Ensure that the `OperationHandler` correctly sets the reason when dispatching a new `SetDirtyStateAction`

- Fix issue in `ChangeBoundsOperationHandler`. Only set the `prefWidth, prefHeight` layout options if the changed object has a `layout` property.
- Cleanup index imports

s21
Fix change-bounds-operation-handler

save model action

s


planger (approved):
Code looks good, thanks! I tested the workflow example server and it seems to still work fine! :+1:

PR #17, opened 2022-04-26
Add `GShapedPrerenderedElement` to graph model

tortmayr:
Add missing `GShapedPrerenderedElement` model element to the graph model.
This can be used to construct special shaped prerendered elements like `ForeignObjectElements`.

Also: Cleanup default type mappings
Requires: https://github.com/eclipse-glsp/glsp-client/pull/178


planger (approved):
I didn't test this as I have no example model, but code-wise it looks fine.
I guess we need a similar change for the `glsp-server` (java) too, right?

tortmayr:
>I guess we need a similar change for the glsp-server (java) too, right?
Yes you are right, I will create an issue to track this.

PR #19, opened 2022-05-04
GH-631 Improve to service bindings in diagram module

tortmayr:
Update behavior of `applyBindingsTarget` for binding `ServiceTargets`.
Add an optional `autoBind` flag to `ServiceTarget` which is enabled by default.
If the target service identifier is not bound yet and this flag is enabled we check
whether the target service can be bound to itself in singleton scope.
If so we do that, otherwise we throw an error.

This covers the most common case for our `ServiceBindings`: Subclassing a default implementation and then bind the service identifier to this implementation.
(e.g. bind(MyModelState).toSelf().inSingletonScope();
bind(ModelState).toService(MyModelState))

For more complex binding configurations its still possible to access the binding context directly (this.context) and apply the bindings this way.
Closes https://github.com/eclipse-glsp/glsp/issues/631
Also: Fix unused model state parameter in workflow builder methods.

### DIFF
--- a/examples/workflow-server/src/handler/create-activity-node-handler.ts
+++ b/examples/workflow-server/src/handler/create-activity-node-handler.ts
@@ -13,7 +13,7 @@
  *
  * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
  ********************************************************************************/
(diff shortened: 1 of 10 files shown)


planger (approved):

PR #21, opened 2022-06-01
Rename `ModelSourceChangedAction` to `SourceModelChangedAction`

planger:
This is to be consistent with the term source model that we now use to refer to the underlying model from which the GModel is created.

Build depends on https://github.com/eclipse-glsp/glsp-client/pull/184

https://github.com/eclipse-glsp/glsp/issues/655


tortmayr (approved):
LGTM

PR #28, opened 2022-12-09
[WIP] #791 Provide  undo-redo feature

tortmayr:
- Rework command statck/command API and implement default command stack
- Provide a generic reusable recording command for JSON objects based on JSON patch
- Implement undo-redo action handler
- Provide a `GModelRecordingCommand` and operation handler to support undo/redo in the gmodel lib 

Breaking
- Remove isDirty flag from `ModelState`.  Dirty state is now managed by the `CommandStack`
- Reworked `CommandStack` and `Command` API.


tortmayr:
For another demo see also https://github.com/eclipse-glsp/glsp-examples/tree/undo-redo-demo
(Note: local linking of the changed node-server package is required to test this)

packages/server-node/src/features/undo-redo/undo-redo-action-handler.ts
```diff
@@ -0,0 +1,57 @@
...
+    @inject(CommandStack)
+    protected commandStack: CommandStack;
+
+    @inject(ModelSubmissionHandler)
+    protected modelSubmissionHandler: ModelSubmissionHandler;
+
+    readonly actionKinds = [UndoOperation.KIND, RedoOperation.KIND];
+
+    execute(action: Action): MaybePromise<Action[]> {
+        if (UndoOperation.is(action)) {
```
tortmayr, line 34: Note: Undo/Redo Action definitions still need to be fixed see https://github.com/eclipse-glsp/glsp/projects/1

packages/server-node/src/gmodel-lib/gmodel-operation-action-handler.ts
```diff
@@ -0,0 +1,56 @@
...
+import * as jsonPatch from 'fast-json-patch';
+import { inject, injectable } from 'inversify';
+import { GModelSerializer } from '../features/model/gmodel-serializer';
+import { ModelState } from '../features/model/model-state';
+import { Command } from '../features/undo-redo/command';
+import { AbstractRecordingCommand } from '../features/undo-redo/recording-command';
+import { OperationActionHandler } from '../operations/operation-action-handler';
+import { OperationHandler } from '../operations/operation-handler';
+@injectable()
+export class GModelOperationActionHandler extends OperationActionHandler {
```
tortmayr, line 28: @planger  I'm not happy with they way this is currently working. The need to rebind the operation action handler just to configure the command seems not optimal to me.
Any ideas/suggestions for improvement?

packages/server-node/src/gmodel-lib/gmodel-operation-action-handler.ts
```diff
@@ -0,0 +1,56 @@
...
+import { Operation, SModelRootSchema } from '@eclipse-glsp/protocol';
+import * as jsonPatch from 'fast-json-patch';
+import { inject, injectable } from 'inversify';
+import { GModelSerializer } from '../features/model/gmodel-serializer';
+import { ModelState } from '../features/model/model-state';
+import { Command } from '../features/undo-redo/command';
+import { AbstractRecordingCommand } from '../features/undo-redo/recording-command';
+import { OperationActionHandler } from '../operations/operation-action-handler';
+import { OperationHandler } from '../operations/operation-handler';
+@injectable()
```
planger, line 27:
```suggestion

@injectable()
```

packages/server-node/src/gmodel-lib/gmodel-operation-action-handler.ts
planger: Hm, yes I see... this is indeed not ideal, also looking at [your example](https://github.com/eclipse-glsp/glsp-examples/commit/1a38dcd4c52c85d31a9b1e7189c8f954aed29485#diff-7a03dee9115e39046295b10ca9eaa8877806bff4d14816defb7521f1b409da5e), the need to rebind a source-model specific operation action handler is a bit unexpected. I'd expect only to bind a source-model-*format*-specific operation action handler (JSON, etc.), which however is generic in the sense of which actual source model (e.g. task list) is used.

Afair, I think in Java this is done differently: there we only have "source-model-format-specific" operation action handlers (e.g. EMF). Thus, if you e.g. have an EMF source model, you'd extend the abstract EMF diagram module, and that's it, because it already binds the EMF-specific operation action handler and you are expected to provide subclasses of [EMFOperationHandler](https://github.com/eclipse-glsp/glsp-server/blob/b500d905454f5cf35ff5a6c3a6b48a61245369a4/plugins/org.eclipse.glsp.server.emf/src/org/eclipse/glsp/server/emf/EMFOperationHandler.java#L28) for your operation handlers. Thus, we delegate more to the operation handlers.

Couldn't the same approach work here too?
An `AbstractJSONOperationHandler` could require the abstract methods `getJsonObject()` and `doExecute()` to be implemented and then keep the OperationActionHandler specific only to JSON or directly to the GModel?

Of course the disadvantage is that every operation handler would need to implement `getJsonObject()`, which is redundant. In customer projects, this was typically avoided by using a common superclass for operation handlers.
So probably this is better than requiring to bind a specific operation action handler? WDYT?

Alternatively, we could be to introduce a new provider that needs to be implemented for JSON-based source models that want to benefit from JSON-patch based undo/redo support. This provider could then return the JSON object from the source model, which should eliminate the need for a specific operation action handler, right?

planger (commented):
Thank you very much, Tobias! This already looks great. I've added a few comments below.

packages/server-node/src/features/model/model-state.ts
```diff
@@ -56,8 +55,6 @@ export class DefaultModelState implements ModelState {
 
     editMode = EditMode.EDITABLE;
 
-    isDirty = false;
```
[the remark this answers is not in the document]
tortmayr: The probelm is even if we keep the `isDirty` property it is essentially a no-op. (As we cannot (and should not) manipulate the command stack internal dirty state from the outside). So even if it doesn't break on a compile level the behavior still breaks.
e.g. `modelState.setDirty= true` no longer works.
I would rather have an obvious API break that is mentioned in the changelog

PR #29, opened 2022-12-09
Start with current element

planger:
I think we should include the current element in the search for an element or parent that matches the guard. This would then be inline with how it is in Sprotty and also better matches the current typedoc (the element to start the search from).

### DIFF
--- a/packages/server-node/src/features/model/gmodel-index.ts
+++ b/packages/server-node/src/features/model/gmodel-index.ts
@@ -177,7 +177,7 @@ export class GModelIndex {
      */
     findParentElement<T extends GModelElement>(elementId: string, guard: TypeGuard<T>): T | undefined {
         const element = this.get(elementId);
-        return element ? this._findElement(element.parent, guard) : undefined;
+        return element ? this._findElement(element, guard) : undefined;
     }
 
     protected _findElement<T>(element: GModelElement, guard: TypeGuard<T>): T | undefined {


tortmayr:
I disagree, this method is meant to be the counter part of the Java Server's [findElementByClass]( https://github.com/eclipse-glsp/glsp-server/blob/a33b31adb6caa7edc65b1072471dfda7a0ca7850/plugins/org.eclipse.glsp.graph/src/org/eclipse/glsp/graph/GModelIndex.java#L162). 
The behavior you implemented does not match the typedoc (it starts with the id of the parent element instead of the given id).
I would rather propose to rename this method to `findElement` or `findElementByType` to align it with the Java Server.

planger:
> I disagree, this method is meant to be the counter part of the Java Server's [findElementByClass](https://github.com/eclipse-glsp/glsp-server/blob/a33b31adb6caa7edc65b1072471dfda7a0ca7850/plugins/org.eclipse.glsp.graph/src/org/eclipse/glsp/graph/GModelIndex.java#L162).

I'm a bit confused... `findElementByClass` in java [_does_](https://github.com/eclipse-glsp/glsp-server/blob/a33b31adb6caa7edc65b1072471dfda7a0ca7850/plugins/org.eclipse.glsp.graph/src/org/eclipse/glsp/graph/GModelIndex.java#L166) include the current element and then walks up to find a match, if the specified element didn't match yet.
My change hence adjusts `findParentElement` to match this behavior. In Sprotty's `findParentByFeature` also includes the current element in the search.

> The behavior you implemented does not match the typedoc (it starts with the id of the parent element instead of the given id).

Not sure I'd read it this way. Imho it doesn't specify whether it includes it or not. Also `findElementByClass` has exactly the same typedoc as the node's `findParentElement`, but `findParentElement` does _not_ include the current element, while `findElementByClass` does.
So imho, if `findParentElement` should be considered the same as `findElementByClass`, it should include the current element (provided as parameter), just as `findElementByClass` does.

Ideally, we should also clarify the typedoc and align their names. I prefer `findParentElement` over `findElementByClass`, as the name already specifies that we walk up the parent hierarchy to find a match instead of, e.g., looking into the children.

tortmayr (approved):
@planger Please disregard my comments. I somehow read the diff in the wrong way and thought you added the parameter from  `element` to `element.parent` 😓. Including the current element is definitely the way to go

planger:
@tortmayr No worries, it is Friday and chances are high in general that I didn't consider something important to begin with. :-) Thanks!

PR #30, opened 2022-12-13
Return port with startup complete message

rschnekenbu:
The complete message on startup now returns the port accepting requests.

If the port is set to 0, OS automatically determines an open port. The port returned by the complete message is the effective port (the specified one if port was set and different of 0, and the effective port if port was set to 0).

Signed-off-by: Remi Schnekenburger [rschnekenburger@eclipsesource.com](mailto:rschnekenburger@eclipsesource.com)

### DIFF
--- a/packages/server-node/src/launch/glsp-server-launcher.ts
+++ b/packages/server-node/src/launch/glsp-server-launcher.ts
@@ -41,7 +41,7 @@ export abstract class GLSPServerLauncher<T = undefined> {
     start(startParams: T): MaybePromise<void> {
         if (!this.running) {
             const result = this.run(startParams);
-            this.running = false;
+            this.running = true;
             return result;
         }
         this.logger.warn('Could not start launcher. Launcher is already running!');
--- a/packages/server-node/src/launch/glsp-server-launcher.ts
+++ b/packages/server-node/src/launch/glsp-server-launcher.ts
@@ -40,8 +40,8 @@ export abstract class GLSPServerLauncher<T = undefined> {
 
     start(startParams: T): MaybePromise<void> {
         if (!this.running) {
-            const result = this.run(startParams);
             this.running = true;
+            const result = this.run(startParams);
             return result;
         }
         this.logger.warn('Could not start launcher. Launcher is already running!');
--- a/packages/server-node/src/launch/socket-server-launcher.ts
+++ b/packages/server-node/src/launch/socket-server-launcher.ts
@@ -19,7 +19,7 @@ import * as jsonrpc from 'vscode-jsonrpc';
 import { GLSPServer, JsonRpcGLSPServer } from '../protocol/glsp-server';
 import { Logger } from '../utils/logger';
 import { GLSPServerLauncher } from './glsp-server-launcher';
-const START_UP_COMPLETE_MSG = '[GLSP-Server]:Startup completed';
+const START_UP_COMPLETE_MSG = '[GLSP-Server]:Startup completed. Accepting requests on port:';
 
 @injectable()
 export class SocketServerLauncher extends GLSPServerLauncher<net.TcpSocketConnectOpts> {
@@ -30,11 +30,24 @@ export class SocketServerLauncher extends GLSPServerLauncher<net.TcpSocketConnec
 
     run(opts: net.TcpSocketConnectOpts): Promise<void> {
         const netServer = net.createServer(socket => this.createClientConnection(socket));
+
         netServer.listen(opts.port, opts.host);
-        this.logger.info(`The GLSP server is ready to accept new client requests on port: ${opts.port}`);
-        // Print a message to the output stream that indicates that the start is completed.
-        // This indicates to the client that the sever process is ready (in an embedded scenario).
-        console.log(this.startupCompleteMessage);
+        netServer.on('listening', () => {
+            const addressInfo = netServer.address();
+            // eslint-disable-next-line no-null/no-null
+            if (addressInfo === null) {
+                this.logger.error('Could not resolve GLSP Server address info');
+                return;
+            } else if (typeof addressInfo === 'string') {
+                this.logger.error('Unexpected type for AddressInfo');
+                return;
+            }
+            const currentPort = addressInfo.port;
+            this.logger.info(`The GLSP server is ready to accept new client requests on port: ${currentPort}`);
+            // Print a message to the output stream that indicates that the start is completed.
+            // This indicates to the client that the server process is ready (in an embedded scenario).
+            console.log(this.startupCompleteMessage.concat(currentPort.toString()));
+        });
         netServer.on('error', () => this.shutdown());
         return new Promise((resolve, reject) => {
             netServer.on('close', () => resolve(undefined));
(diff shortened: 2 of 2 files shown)


packages/server-node/src/launch/socket-server-launcher.ts
```diff
@@ -30,11 +30,24 @@ export class SocketServerLauncher extends GLSPServerLauncher<net.TcpSocketConnec
...
-        this.logger.info(`The GLSP server is ready to accept new client requests on port: ${opts.port}`);
-        // Print a message to the output stream that indicates that the start is completed.
-        // This indicates to the client that the sever process is ready (in an embedded scenario).
-        console.log(this.startupCompleteMessage);
+        netServer.on('listening', () => {
+            const addressInfo = netServer.address();
+            // eslint-disable-next-line no-null/no-null
+            if (addressInfo === null) {
+                this.logger.error('Could not resolve GLSP Server address info');
+                return;
```
sdirix, line 40: We should probably call `this.shutdown()` as we don't expect anyone to connect in this case

packages/server-node/src/launch/socket-server-launcher.ts
```diff
@@ -30,11 +30,24 @@ export class SocketServerLauncher extends GLSPServerLauncher<net.TcpSocketConnec
...
+            } else if (typeof addressInfo === 'string') {
+                this.logger.error('Unexpected type for AddressInfo');
+                return;
```
sdirix, line 43: same here

packages/server-node/src/launch/socket-server-launcher.ts
```diff
@@ -30,11 +30,24 @@ export class SocketServerLauncher extends GLSPServerLauncher<net.TcpSocketConnec
...
         netServer.listen(opts.port, opts.host);
-        this.logger.info(`The GLSP server is ready to accept new client requests on port: ${opts.port}`);
-        // Print a message to the output stream that indicates that the start is completed.
-        // This indicates to the client that the sever process is ready (in an embedded scenario).
-        console.log(this.startupCompleteMessage);
+        netServer.on('listening', () => {
+            const addressInfo = netServer.address();
+            // eslint-disable-next-line no-null/no-null
+            if (addressInfo === null) {
+                this.logger.error('Could not resolve GLSP Server address info');
```
sdirix, line 39:
```suggestion
                this.logger.error('Could not resolve GLSP Server address info. Shutting down.');
```

packages/server-node/src/launch/socket-server-launcher.ts
```diff
@@ -30,11 +30,24 @@ export class SocketServerLauncher extends GLSPServerLauncher<net.TcpSocketConnec
...
+                return;
+            } else if (typeof addressInfo === 'string') {
+                this.logger.error('Unexpected type for AddressInfo');
```
sdirix, line 42: This should log more detailed information: 
```suggestion
                this.logger.error(`GLSP Server is unexpectedly listening to pipe or domain socket "${addressInfo}". Shutting down.`);
```

sdirix (changes requested):
Looks good!

packages/server-node/src/launch/socket-server-launcher.ts
```diff
@@ -30,11 +30,24 @@ export class SocketServerLauncher extends GLSPServerLauncher<net.TcpSocketConnec
...
+
         netServer.listen(opts.port, opts.host);
-        this.logger.info(`The GLSP server is ready to accept new client requests on port: ${opts.port}`);
-        // Print a message to the output stream that indicates that the start is completed.
-        // This indicates to the client that the sever process is ready (in an embedded scenario).
-        console.log(this.startupCompleteMessage);
+        netServer.on('listening', () => {
+            const addressInfo = netServer.address();
+            // eslint-disable-next-line no-null/no-null
+            if (addressInfo === null) {
```
tortmayr, line 38: You can  avoid the null check and eslint supression by using this:
```suggestion
            if (!addressInfo) {
```

tortmayr (changes requested):

packages/server-node/src/launch/socket-server-launcher.ts
rschnekenbu: I get this message when I add `this.shutdown()`: _Could not stop launcher. Launcher is not running!_
It seems for me that in run(), [this.running](https://github.com/eclipse-glsp/glsp-server-node/blob/35b3b9dff6b49707605c479827cb9debbed3079f/packages/server-node/src/launch/glsp-server-launcher.ts#L44) should be set to 'true'.

packages/server-node/src/launch/socket-server-launcher.ts
rschnekenbu: Thanks @tortmayr ! Changed in current commit

packages/server-node/src/launch/socket-server-launcher.ts
rschnekenbu: Thanks @sdirix, I updated the log info with details

packages/server-node/src/launch/socket-server-launcher.ts
rschnekenbu: done

packages/server-node/src/launch/socket-server-launcher.ts
rschnekenbu: same comment as https://github.com/eclipse-glsp/glsp-server-node/pull/30#discussion_r1047541781 applies here.

packages/server-node/src/launch/socket-server-launcher.ts
rschnekenbu: To test that, I ran the "Debug Workflow Example Server" target, with a breakpoint on line 36 of packages/server-node/src/launch/socket-server-launcher.ts, e.g. `const addressInfo = netServer.address();`
If adressInfo is changed to a string, we can run into the else if [String] branch and test shutdown.

packages/server-node/src/launch/glsp-server-launcher.ts
```diff
@@ -41,7 +41,7 @@ export abstract class GLSPServerLauncher<T = undefined> {
     start(startParams: T): MaybePromise<void> {
         if (!this.running) {
             const result = this.run(startParams);
-            this.running = false;
+            this.running = true;
```
tortmayr, line 44: We should set running to `true` before we call `this.run`.
Otherwise calling `shutdown()`  will fail in non-async run implementations;

tortmayr (changes requested):
Thanks for the fast update @rschnekenbu. Code looks good
to me, I only noticed one minor issue

packages/server-node/src/launch/glsp-server-launcher.ts
rschnekenbu: Good catch! updated with next commit.

packages/server-node/src/launch/socket-server-launcher.ts
rschnekenbu: This has been updated in next commit (invert  order of running).

rschnekenbu:
@tortmayr, all comments are adressed now. Thanks for reviewing :)

tortmayr (approved):
Thanks! LGTM 👍🏼

PR #33, opened 2022-12-21
Fix `stop` of socket server

eneufeld:
The socket server currently does not actually shutdown the server when calling shutdown.
In order to do so, the `close` method must be called on the server. Thus the server was moved into a private member
to be able to call `close` on `stop`.
This also fixes the test launch config and adds a test. Furthermore the overriden methods in SocketServerLauncher are not public anymore in order to have a clean api.

Contributed on behalf of STMicroelectronics.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -6,7 +6,7 @@
       "type": "node",
       "request": "launch",
       "program": "${workspaceRoot}/node_modules/.bin/mocha",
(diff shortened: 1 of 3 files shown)


tortmayr (changes requested):
Thanks @eneufeld! Change looks good to me and the server shutdown now seem to properly dispose all resources 👍🏼  I have a few minor remarks/nitpicks as inline comments.

tortmayr (approved):
LGTM! Thanks

PR #35, opened 2023-01-03
Fix issues with copy&paste feature

tortmayr:
### Fix issues with copy&paste feature

- Fix incorrect edge rewiring in `CopyPasteOperationHandler`
- Fix incorrect usage of assignment (=) instead of comparison (===) in   typeguard functions of graph package
- Update feature matrix in README.md

### Update glsp-config
Update to latest glsp-config version & autoformat package.json files 
Add check:pr & check:header scripts


martin-fleck-at (approved):
Change looks good to me! Tested both the copy/paste and the navigation, thank you Tobi!

PR #36, opened 2023-01-18
Ensure that all services that return MaybePromises are properly awaited

tortmayr:
Otherwise we the server might crash in error cases due to unhandled promise rejections

Fixes https://github.com/eclipse-glsp/glsp/issues/868

General cleanups/minor refactorings:
- Remove Disposable and MaybePromise API and use protocol definitions instead.
- Ensure that Action/Operation handlers use MaybePromise as return type when applicable
- Ensure that the GLSPServer.process method forwards all unhandled errors to the client as notification
- 

Requires https://github.com/eclipse-glsp/glsp-client/pull/222


martin-fleck-at (approved):
Change looks good to me and works well. Thank you very much Tobias!

PR #37, opened 2023-01-26
Refactor packages and provide dedicated browser and node runtimes

tortmayr:
- Move shared code into common package, add dedicated browser and node packages
- Update package.json to provide entry points for both node and browser
- Refactor `@eclipse-glsp-examples/workflow-server` following the same pattern
- Add webpack config to enable both bundling for node and the browser
- Update changelog

- Add bundle step for browser-env to Jenkins build

Fixes https://github.com/eclipse-glsp/glsp/issues/884


lerna.json
```diff
@@ -10,7 +10,7 @@
     "publish": {
       "forcePublish": true,
       "skipGit": true,
-      "registry": "https://registry.npmjs.org/"
+      "registry": "http://localhost:4873/"
```
xai, line 13: Is this change intentional?

xai (changes requested):
Looks good and works great with glsp-theia-integration and glsp-vscode-integration :+1: 
I also tested debugging the workflow example and the node backend.

Really minor comments:
- The main [README](https://github.com/eclipse-glsp/glsp-server-node/blob/614e5fdb5a30d635ae9776b80341eff01cdfefb9/README.md?plain=1#L8) still refers to `@eclipse-glsp/server-node`
- We could hint to the minimum nodejs requirement (16) in the README and maybe even include a .nvmrc to suggest a tested version of node to the user.

lerna.json
tortmayr: Good catch. No that chage was on accident and is a remainder from local testing 😅

tortmayr:
Thanks Olaf, I have updated the README and fixed the `lerna.json` file

> * We could hint to the minimum nodejs requirement (16) in the README and maybe even include a .nvmrc to suggest a tested version of node to the user.
Sounds like a good plan. However, if we do that we should do it consistently across all repositories. So we should probably track and handle this in a follow-up issue

xai (approved):
LGTM!

PR #38, opened 2023-01-30
Provide command API

tortmayr:
### Implement command API & Undo/Redo support
- implement a generic Command/Commandstack API similar to the EMF API
- Provide default implementations for Commandstack & Compound command
- Provide a RecordingCommand API for arbitary JSON models using json patches
- Add testscases for the new command API 
Fixes https://github.com/eclipse-glsp/glsp/issues/791

### Refactor Operationhandler API (breaking change)
- Refactor OperationHandler API to return optional executable commands (similar to how it has been done for the Java API)
- Unfortunately this introduces hard breaks and maintaining a compatibility/deprecation layer is not easily feasible.
- Most prominent changes: Refactor `OperationHandler` from interface to class and refactor `CreateOperationHandler` to interface to facilitate multi-inheritance 
Fixes https://github.com/eclipse-glsp/glsp/issues/889

### Ensure that all components of the direct gmodel library are correctly prefixed. 
Fixes https://github.com/eclipse-glsp/glsp/issues/826

### Implement direct editor feature for the workflow example


packages/server/src/common/operations/compound-operation-handler.ts
```diff
@@ -15,20 +15,29 @@
...
 @injectable()
-export class CompoundOperationHandler implements OperationHandler {
+export class CompoundOperationHandler extends OperationHandler {
+    @inject(OperationHandlerRegistry)
+    protected operationHandlerRegistry: OperationHandlerRegistry;
+
     operationType = CompoundOperation.KIND;
 
-    @inject(OperationHandlerRegistry) protected operationHandlerRegistry: OperationHandlerRegistry;
+    async createCommand(operation: CompoundOperation): Promise<Command | undefined> {
```
martin-fleck-at, line 30: Since we are no longer calling `executeNestedOperation` I believe we can simply remove that method.

packages/server/src/common/command/command-stack.ts
```diff
@@ -13,62 +13,131 @@
...
 
-    isDirty(): boolean {
-        return false;
+    get isDirty(): boolean {
+        return this.undoStack.length > 0;
     }
 
     saveIsDone(): void {
-        // no-op
+        this.flush();
```
martin-fleck-at, line 129: This behavior does not match the Java side as with the `flush` we can no longer do any `undo` after the save which is a bit unexpected. In Java/EMF, the behavior is to simply remember the last save index and compare it to the index of the last executed operation, i.e., it is more a shifting of pointers in the command history. I think it would be nice to match this behavior here as well, what do you think?

martin-fleck-at (changes requested):
Hi @tortmayr, it's hard to express how much I like this change. I really appreciate the consistency with the Java side, the naming cleanup. As far as I can see from my tests, everything works as expected, there is only one unexpected behavior that I experienced when saving as it prevents me from undoing any changes I previously made (see also my comment inline). Other than that, it is really fantastic work, thank you very much!

packages/server/src/common/operations/compound-operation-handler.ts
tortmayr: Good point. Is removed with 41f9759

packages/server/src/common/command/command-stack.ts
tortmayr: Good catch. I have updated to commandstack implementation to be inline with the EMF implementation (41f9759)

packages/server/src/common/command/command-stack.ts
```diff
@@ -13,62 +13,150 @@
...
-        // no-op
+        if (this.canRedo()) {
+            const command = this.commands[++this.top];
+            try {
+                command.redo();
+            } catch (error) {
+                this.handleError(error);
+
+                // Clear the command list past the top (i.e. the old redo list)
+                this.commands = this.commands.slice(0, this.top + 1);
```
martin-fleck-at, line 133: This is unreachable code as the `handleError` method will always throw an error. Also it seems to me that the handleError will always flush the whole command stack so there is no point in slicing anything.

martin-fleck-at (changes requested):
Thank you very much for the quick update! The command stack now seems to work as expected, very cool! However, there are some code oddities which we should fix before merging this change.

packages/server/src/common/command/command-stack.ts
tortmayr: Good catch, no need to slice in the redo method.
In the execute method, however, we need the slicing because after successful execution of a new command we have to clear all "old" redo commands.

martin-fleck-at: Got it! :+1:

martin-fleck-at (approved):
Fantastic work, everything is working as expected for me, thank you very much Tobias!

PR #39, opened 2023-02-28
Support async command execution

tortmayr:
Change command API to support `MaybePromise<void>` instead of just plain voids. This enables async command execution e.g. for communicating with a modelserver 

### DIFF
--- a/packages/server/src/common/command/command.ts
+++ b/packages/server/src/common/command/command.ts
@@ -106,17 +106,17 @@ export class CompoundCommand implements Command {
         }
     }
 
-    redo(): void {
+    async redo(): Promise<void> {
         const alreadyRedone: Command[] = [];
 
         try {
             for (const command of this.commands) {
-                command.redo();
+                await command.redo();
                 alreadyRedone.unshift(command);
             }
         } catch (err) {
             for (const command of alreadyRedone) {
-                command.undo();
+                await command.undo();
             }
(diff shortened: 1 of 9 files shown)


packages/server/src/common/command/command.ts
```diff
@@ -108,12 +110,14 @@ export class CompoundCommand implements Command {
         const alreadyRedone: Command[] = [];
 
         try {
-            this.commands.forEach(command => {
+            for (const command of this.commands) {
```
martin-fleck-at, line 113: Should we not also await the `redo` in the line below to make sure that the order is properly kept and there are no conflicts?

packages/server/src/common/command/command.ts
```diff
@@ -108,12 +110,14 @@ export class CompoundCommand implements Command {
...
                 command.redo();
                 alreadyRedone.unshift(command);
-            });
+            }
         } catch (err) {
-            alreadyRedone.forEach(command => command.undo());
+            for (const command of alreadyRedone) {
+                command.undo();
```
martin-fleck-at, line 119: Should we not also await the `undo` here to make sure that the order is properly kept and there are no problems?

martin-fleck-at (changes requested):
Change looks good to me for the most part, thank you very much Tobias! I just have a question regarding the compound command.

PR #40, opened 2023-03-03
Ensure that recording commands can be executed asynchronous

tortmayr:
Change signature of `applyPatch` and `getJsonObject` to also  allow promise return values 
Also: add missing exports to main index


martin-fleck-at (approved):
LGTM!

PR #42, opened 2023-04-06
feat: adjust server startup to use autoassigned ports

xai:
This sets the default port to 0, which will cause the server to automatically assign a port. The port is then printed to the console in the startup message and can be parsed by clients.

Part of eclipse-glsp/glsp#965

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@
         -   Refactored `CreateOperationHandler` to an interface instead of a class
         -   Renamed the services and handlers of the direct GModel library => consistent use of `GModel` prefix
         -   The `ModelState` interface no longer has an `isDirty` flag. Dirty state is now handled by the `CommandStack`
+-   [server] Default port has changed from 5007 (and 8081 for websocket) to 0, which implies autoassignment by the OS [#42](https://github.com/eclipse-glsp/glsp-server-node/pull/42)
 
 ## [v1.0.0 - 30/06/2022](https://github.com/eclipse-glsp/glsp-server-node/releases/tag/v1.0.0)
 
--- a/examples/workflow-server/package.json
+++ b/examples/workflow-server/package.json
@@ -49,7 +49,7 @@
     "lint": "eslint --ext .ts,.tsx ./src",
     "lint:ci": "yarn lint -o eslint.xml -f checkstyle",
     "prepare": "yarn clean && yarn build",
-    "start": "node --enable-source-maps bundle/wf-glsp-server-node.js",
+    "start": "node --enable-source-maps bundle/wf-glsp-server-node.js --port 5007",
     "start:websocket": "node --enable-source-maps bundle/wf-glsp-server-node.js -w --port 8081",
     "watch": "tsc -w"
   },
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -27,6 +27,8 @@
         -   Refactored `CreateOperationHandler` to an interface instead of a class
         -   Renamed the services and handlers of the direct GModel library => consistent use of `GModel` prefix
         -   The `ModelState` interface no longer has an `isDirty` flag. Dirty state is now handled by the `CommandStack`
+-   [server] Default port has changed from 5007 to 0, which implies autoassignment by the OS
```
tortmayr, line 30: It don`t think it's necessary to split this into bullet points. One pullet points covering both should be enough. Please also adda link to the PR at the end like this: [#42]](https://github.com/eclipse-glsp/glsp-server-node/pull/42/)

tortmayr (approved):
LGTM. Only one minor nitpick regarding the changelog

README.md
```diff
@@ -61,7 +61,7 @@ To launch the server for TCP sockets use:
 yarn start
 ```
 
-This starts a server that is listening on port 5007 for incoming client requests.
+This starts a server that is listening on port 0 for incoming client requests.
```
tortmayr, line 64: Should still start  at port 5007. The server cannot be tested standalone, so it has always be tested with an integration project in "external server mode". For this we need an explicit port to connect to.

README.md
```diff
@@ -75,7 +75,7 @@ To launch the server for WebSockets use:
 yarn start:websocket
 ```
 
-This starts a server that is listening on the `ws://localhost:8081/workflow` endpoint for incoming client requests.
+This starts a server that is listening on the `ws://localhost:0/workflow` endpoint for incoming client requests.
```
tortmayr, line 78: Same has above, we should still use an explicit port here.

examples/workflow-server/package.json
```diff
@@ -50,7 +50,7 @@
     "lint:ci": "yarn lint -o eslint.xml -f checkstyle",
     "prepare": "yarn clean && yarn build",
     "start": "node --enable-source-maps bundle/wf-glsp-server-node.js",
-    "start:websocket": "node --enable-source-maps bundle/wf-glsp-server-node.js -w --port 8081",
+    "start:websocket": "node --enable-source-maps bundle/wf-glsp-server-node.js -w --port 0",
```
tortmayr, line 53:
```suggestion
    "start:websocket": "node --enable-source-maps bundle/wf-glsp-server-node.js -w --port 8081",
```

examples/workflow-server/package.json
```diff
@@ -50,7 +50,7 @@
     "lint:ci": "yarn lint -o eslint.xml -f checkstyle",
     "prepare": "yarn clean && yarn build",
     "start": "node --enable-source-maps bundle/wf-glsp-server-node.js",
```
tortmayr, line 52:
```suggestion
    "start": "node --enable-source-maps bundle/wf-glsp-server-node.js --port 5007",
```

tortmayr (changes requested):

tortmayr (approved):
LGTM! Thanks

PR #43, opened 2023-04-24
Distinguish between live and batch validation

planger:
https://github.com/eclipse-glsp/glsp/issues/980

Corresponding change for Java server: https://github.com/eclipse-glsp/glsp-server/pull/200

### DIFF
--- a/packages/server/src/common/features/model/model-submission-handler.ts
+++ b/packages/server/src/common/features/model/model-submission-handler.ts
@@ -17,16 +17,20 @@ import { GModelRootSchema } from '@eclipse-glsp/graph';
 import {
     Action,
     DirtyStateChangeReason,
+    MarkersReason,
     MaybePromise,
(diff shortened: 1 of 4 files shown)


packages/server/src/common/features/model/model-submission-handler.ts
```diff
@@ -52,6 +56,10 @@ export class ModelSubmissionHandler {
     @inject(CommandStack)
     protected commandStack: CommandStack;
 
+    @inject(ModelValidator)
+    @optional()
+    protected validator: ModelValidator;
```
tortmayr, line 61: An optional inject should also be defined as optional type:
```suggestion
    protected validator?: ModelValidator;
```

tortmayr (approved):
Changes look good to me.
I only have to minor nitpicks, that could be improved but do not prevent merging

PR #44, opened 2023-04-24
GLSP-975 Ensure generic GLSP Server implementation

tortmayr:
- Introduce a common subclass for jsonrpc based ServerLaunchers
- Move out jsonrpc implementation details from `GLSPServer` implementation into the `JsonRpcGLSPServerLauncher`. This ensures that the Server implementation itself is independent from the chosen communication channel and can reused for other use cases like direct invocation in a node backend or orthogonal communication channels like REST or gRPC.
- Update to latest @eclipse-glsp/protocol version and consume shared server interface types from there

- Ensure that the `@eclipse-glsp-examples/workflow-server` package is properly indexed
- Add entry points for browser and node 
- Move rpc-server initialization out of index files into app files. 

Also: Update example applications to ensure that promise rejections/errors are properly catched.

Part of https://github.com/eclipse-glsp/glsp/issues/975

Requires https://github.com/eclipse-glsp/glsp-client/pull/245


packages/server/src/common/utils/console-logger.ts
sgraband, line 1: I guess this should replace `packages/server/src/browser/di/console-logger`?
The old file however was not removed as far as i can see. Could you do that?

packages/server/src/common/session/client-session-listener.ts
```diff
@@ -13,6 +13,7 @@
  *
  * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
  ********************************************************************************/
+
```
sgraband, line 16: Minor, but this just introduces an empty line, which is not required by the linter afaik.

examples/workflow-server/browser.d.ts
sgraband, line 1: Should this file be checked in?
If yes, could we maybe add a comment for why this is needed?

examples/workflow-server/browser.js
sgraband, line 1: Should this file be checked in?
If yes, could we maybe add a comment for why this is needed?

examples/workflow-server/node.d.ts
sgraband, line 1: Should this file be checked in?
If yes, could we maybe add a comment for why this is needed?

examples/workflow-server/node.js
sgraband, line 1: Should this file be checked in?
If yes, could we maybe add a comment for why this is needed?

sgraband (commented):
Thank you for the changes! They look good and work great! I just have a few questions/comments inline below. Could you take a look at them?

packages/server/src/common/utils/console-logger.ts
tortmayr: Good catch. Yes exactly, the intention was to move the generic `ConsoleLogger` from the browser into the common package. I will remove the redundant file.

examples/workflow-server/browser.d.ts
sgraband: Makes sense. Thank you for the explanation. I was simply confused by the location of the files, but i don't think we need a comment then.

PR #47, opened 2023-05-22
GLSP-1004 Update publish process

tortmayr:
Exclude spec file from published content.
Also include bundled browser workflow  sever into published content

Part of https://github.com/eclipse-glsp/glsp/issues/1004


martin-fleck-at (approved):
LGTM!

martin-fleck-at (approved):
LGTM!

PR #48, opened 2023-05-27
GLSP-1022: Update to TS5 and inversify 6

tortmayr:
- Update to typescript 5
- Update to inversify 6. Inversify is now a peerDependency (^6.0.0). A matching version has to be provided by consuming projects.
- Adapt codebase to new inversify API
- Rework elk-layouit package to remove unnecessary dependency in `sprotty-elk`
Part of https://github.com/eclipse-glsp/glsp/issues/1022
(uses locally published glsp packages. needs to be adapted once the offical packages are available)

Requires https://github.com/eclipse-glsp/glsp-client/pull/253


packages/server/src/common/utils/args-util.ts
```diff
@@ -48,7 +48,7 @@ export namespace ArgsUtil {
         }
 
         return {
-            [KEY_RADIUS_TOP_RIGHT]: topLeft,
+            [KEY_RADIUS_TOP_LEFT]: topLeft,
```
martin-fleck-at, line 51: Nice catch!

.eslintrc.js
```diff
@@ -9,13 +9,13 @@ module.exports = {
...
     rules: {
-        'no-shadow': 'off',
+        '@typescript-eslint/no-shadow': 'off',
         'no-restricted-imports': [
             'warn',
             {
                 name: 'sprotty-protocol',
                 message:
-                    "The sprotty-protocol default exports are customized and reexported by GLSP. Please import from '@eclipse-glsp/client' instead"
+                    "The sprotty-protocol default exports are customized and reexported by GLSP. Please import from '@eclipse-glsp/protocol' instead"
```
martin-fleck-at, line 18: Is that true? It seems to me that even with https://github.com/eclipse-glsp/glsp-client/pull/253 we still re-export the sprotty types from '@eclipse-glsp/client'?

martin-fleck-at (changes requested):
Looks good to me in general. However, the adaptation of the linting message seems odd to me, please double check.

.eslintrc.js
tortmayr: The `@eclipse-glsp/protocol` package has a partial reexport of the needed `sprotty-protocl` tpyes (https://github.com/eclipse-glsp/glsp-client/blob/d81a1ea5335410d2daeaa0b869ed7573eaff6471/packages/protocol/src/index.ts#L43). There should be no reason to ever directly use the `sprotty-protocol` package in a GLSP context.

martin-fleck-at: I understand that we don't want `sprotty-protocol`. I was just confused because in all other linting rules we suggest to use `@eclipse-glsp/client` instead but here it is `@eclipse-glsp/protocol` but it seems that there is a re-export in both, the client and in the protocol package and I the server should not have a dependency on the client ;-)

martin-fleck-at (approved):
LGTM!

PR #49, opened 2023-06-06
GLSP-1026 Provide bundle on example server package

tortmayr:
Introduce a `@eclipse-glsp-examples/workflow-server-bundle` package. This package just contains the bundled workflowserver without any direct dependencies to server components. It can be consumed in the dev example of the client repository without introducing a dependecy loop between client & server. 

This fixes eclipse-glsp/glsp/issues/1026. We can decide on a later point if want to move the protocol package into the umbrella repository. In regards to developer experience its probably better to keep it in the client repo. Changes to the protocol package typically also  require changes in the client and server repos. By keeping the package in one of the repositories in question we can reduce the required overhead with regards to merging and consuming the changes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,4 +14,5 @@ tsconfig.tsbuildinfo
 eslint.xml
 report.xml
 
-examples/workflow-server/bundle
+wf-glsp-server-node.js
+wf-glsp-server-node.js.map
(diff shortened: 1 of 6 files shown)


martin-fleck-at (approved):
LGTM!

PR #54, opened 2023-07-20
GLSP-77: Allow WebSocket connections to reconnect after interrupt

ndoschek:
If Websocket connections reconnect after an interrupt, ensure that the last modelState is restored on RequestModelAction if available

Part of https://github.com/eclipse-glsp/glsp/issues/77

---

Related to: https://github.com/eclipse-glsp/glsp-client/pull/269

### DIFF
--- a/packages/server/src/common/features/model/request-model-action-handler.ts
+++ b/packages/server/src/common/features/model/request-model-action-handler.ts
@@ -17,6 +17,7 @@ import { Action, RequestModelAction, ServerStatusAction } from '@eclipse-glsp/pr
 import { inject, injectable } from 'inversify';
 import { ActionDispatcher } from '../../actions/action-dispatcher';
 import { ActionHandler } from '../../actions/action-handler';
+import { ClientOptionsUtil } from '../../utils/client-options-util';
 import { Logger } from '../../utils/logger';
 import { ProgressMonitor, ProgressService } from '../progress/progress-service';
 import { ModelState } from './model-state';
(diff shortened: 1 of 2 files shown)


packages/server/src/common/features/model/request-model-action-handler.ts
```diff
@@ -49,8 +50,24 @@ export class RequestModelActionHandler implements ActionHandler {
         this.logger.debug('Execute RequestModelAction:', action);
         this.modelState.setAll(action.options ?? {});
 
+        const isReconnecting = ClientOptionsUtil.isReconnecting(action.options);
+
         const progress = this.reportModelLoading('Model loading in progress');
-        await this.sourceModelStorage.loadSourceModel(action);
+
+        if (isReconnecting) {
+            const oldModelRoot = this.modelState.root;
```
tortmayr, line 58: Please extract the reconnect functionality in a protected submethod so that it's easier for adopters to customize the behavior.

packages/server/src/common/features/model/request-model-action-handler.ts
```diff
@@ -49,8 +50,24 @@ export class RequestModelActionHandler implements ActionHandler {
...
+            if (oldModelRoot) {
+                // use current modelRoot of modelState and submit
+                this.modelState.updateRoot(oldModelRoot);
```
tortmayr, line 61: Why is this necessary? We are basically resuming where we left of before the reconnect. So the modelState and indexing should already  be up to date.

tortmayr (changes requested):
Looks good to me! I only have a few minor suggestions in line

packages/server/src/common/features/model/request-model-action-handler.ts
ndoschek: Thanks, that's true, I removed that line again!

tortmayr (approved):

PR #55, opened 2023-09-04
GLSP-1101: Fix Windows path  conversion

tortmayr:
Ensure leading shashes in File urls are properly handled under Windows. 
Also: Fix rimraf command that was not working under Windows 
Fixes https://github.com/eclipse-glsp/glsp/issues/1101


planger (approved):

Can't test this directly, but looks good code-wise.

PR #56, opened 2023-09-04
GLSP-1109: Remove dependency to fs-extra

sdirix:
Use 'fs' instead of 'fs-extra' in GLSP server node as the additional capatibilities of 'fs-extra' are not used. Also fixes the issue that the dependency to 'fs-extra' was not declared in the 'package.json' which might fail consumer builds.

Contributed on behalf of STMicroelectronics.

### DIFF
--- a/packages/server/package.json
+++ b/packages/server/package.json
@@ -68,7 +68,6 @@
     "ws": "^8.12.1"
   },
   "devDependencies": {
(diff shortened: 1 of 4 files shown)


tortmayr (approved):
LGTM! Thanks! 👍🏼

PR #57, opened 2023-09-15
Adapt to latest protocol changes

tortmayr:
# GLSP-1116 Revise model loading
- Refactor `ModelSubmissionHandler` to support proper handling of the `RequestModelAction` as real request action 
Part-of: https://github.com/eclipse-glsp/glsp/issues/1116

# GLSP-1117: Remove need for explicit definition of client actions

Refactor the base GLSP protocol to allow the client to tell the server which actions it is going to handle i.e. which actions should be forwarded to the client
- Add `clientActions` array to `InitializeClientSessionParams`. This means the client now has to pass the action kinds it wants to handle as part of the initalize request
- Replace `ClientActionHandler` with `ClientActionForwader` a separate component that is not part of the server-side action handlers. 
- Remove `configureClientActions` method from `DiagramModule` as the explicit configuration is no longer needed

Part of eclipse-glsp/glsp/issues/1117

# GLSP-1071: Rename ServerStatus/ServerMessage action Part of eclipse-glsp/glsp/issues/1071

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -21,7 +21,7 @@
       "args": [
         "--config",
         "${workspaceFolder}/.mocharc",
-        "${workspaceFolder}/packages/server-node/src/**/*.spec.ts"
+        "${workspaceFolder}/packages/server/src/**/*.spec.ts"
       ],
       "console": "integratedTerminal",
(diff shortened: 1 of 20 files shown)


planger (approved):

PR #60, opened 2023-09-26
GLSP-211: Revise TypeHints and server side feedback for creation acti…

tortmayr:
…ons (#210)

- Add action handler for `RequestEdgeCheckAction` and `EdgeCreationChecker` API
- Add a optional `EdgeCreationchecker` component that can be implemented by adopters to provide dynamic typehints
- Adapt workflow example to use dynamic edge hints for weighted edges

Also: Update eslint config to show warnings if we import from internal index instead of the relative path to the file

Fixes eclipse-glsp/glsp/issues/211


tortmayr:
Good catch. I forgot to upgrade the protocol version in the yarn.lock.

examples/workflow-server/src/common/workflow-edge-creation-checker.ts
```diff
@@ -0,0 +1,31 @@
...
+ ********************************************************************************/
+import { EdgeCreationChecker, GModelElement } from '@eclipse-glsp/server/';
+import { injectable } from 'inversify';
+import { TaskNode } from './graph-extension';
+import { ModelTypes } from './util/model-types';
+
+@injectable()
+export class WorkflowEdgeCreationChecker implements EdgeCreationChecker {
+    isValidSource(edgeType: string, sourceElement: GModelElement): boolean {
+        return edgeType === ModelTypes.WEIGHTED_EDGE;
```
martin-fleck-at, line 24: nit: In the Java variant we also have this condition: `sourceElement.getType().equals(DECISION_NODE)`.

martin-fleck-at (approved):
LGTM! Minor nitpick that you may or may not consider.

PR #62, opened 2023-10-13
Bugfixes for 2.0

tortmayr:
- Fix cut operation handler Ensure that the GModelCutOperation handler does not execute a command itself and instead dispatches a corresponding DeleteOperation
- Fix dependecy resolvement in delete-operaetion handler


martin-fleck-at (approved):
LGTM!

PR #67, opened 2023-11-14
Use '127.0.0.1' instead of 'localhost' to avoid ipv4/ipv6 issues

martin-fleck-at:
Fixes https://github.com/eclipse-glsp/glsp/issues/1164


tortmayr (approved):
LGTM!  Thanks

PR #68, opened 2023-11-16
GLSP-1163: Fix GModelCreateNodeOperationHandler

tortmayr:
- Ensure that the element to delete is removed from its direct parent instead of always using the model root
- Remove lookup of topLevel node or edge before deletion.
- Ensure that edges of ports are also considered during dependants collection
- Avoid adding of child elements to the dependants collection (as they are implicitly deleted when deleting the parent)
- Fix to narrow typing in GModelIndex.getIncomingEdges

Fixes https://github.com/eclipse-glsp/glsp/issues/1163

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -13,7 +13,6 @@
     "javascript",
     "typescript"
   ],
-  "prettier.prettierPath": "node_modules/prettier",
   "search.exclude": {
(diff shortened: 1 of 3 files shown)


martin-fleck-at (approved):
Works really well, even if we nest several categories and nodes, thank you Tobias!

PR #69, opened 2023-11-26
Ensure async usage of JsonOperationHandler.commandOf

tortmayr:
Make runnable of command return a MaybePromise to be consistent with the doExecute method of recording commands and support async recordings.


martin-fleck-at (approved):
LGTM!

PR #74, opened 2024-01-31
GLSP-1227: Ensure that model update is correct when using autolayout

tortmayr:
In the `submmitModelDirectly` method of the `ModelSubmissionHandler` the gmodel serialization was invoked before the layout engine was executed (in automatic layouting mode). As a consequence the model sent to the server reflected the unlayouted state.

Fixes https://github.com/eclipse-glsp/glsp/issues/1227


martin-fleck-at (approved):
LGTM!

PR #76, opened 2024-02-21
GLSP-1266 Fix handling of zero values in builder functions

ndoschek:
#### What it does

- Fix handling of zero values in `GBoundsAwareBuilder` functions
- Fix handling of zero values in `GAlignableBuilder` function
- Adds unit tests for those updated functions

Resolves eclipse-glsp/glsp/issues/1266

#### Changelog

### DIFF
--- a/packages/graph/src/galignable.spec.ts
+++ b/packages/graph/src/galignable.spec.ts
@@ -1,0 +1,74 @@
+/********************************************************************************
+ * Copyright (c) 2024 EclipseSource and others.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License v. 2.0 which is available at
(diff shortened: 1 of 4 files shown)


tortmayr (changes requested):
Looks good to me. Could you please also make the same adjustment for the `GAlignableBuilder` functions?
Also please update the `Resolves` comment in your commit to `Resolves  eclipse-glsp/glsp/issues/1266` otherwise autoclosing on merge won't work.

ndoschek:
Sure, thanks for the review @tortmayr!

ndoschek:
I found some motivation and added a few unit tests for the functions I updated, would be great if you could have a quick look on those as well when you get a chance. Thanks!

tortmayr (approved):
Thanks. More test coverage is always appreciated 😉

PR #78, opened 2024-04-06
Fixed webworker startup bc elkjs hidden dependency

Sakrafux:
As described in the comment, but `elkjs` wants to instantiate workers. If no `workerFactory` is given, it tries to resolve this using a fake worker from `elk-worker.min.js`, which is seemingly available when started through node, but not in a web-worker context. Since it is a dynamic import unknown to webpack, it is not bundled.

Therefore, I simply assigned a simple stub, since the worker functionality of `elkjs` is not needed. This enables the server to start in a worker as expected.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,5 @@ report.xml
 
 wf-glsp-server-node.js
 wf-glsp-server-node.js.map
+wf-glsp-server-webworker.js
+wf-glsp-server-webworker.js.map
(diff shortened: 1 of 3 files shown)


tortmayr (approved):
Good catch! 
Changes look good to me. Thanks! 👍🏼 
This also fixes https://github.com/eclipse-glsp/glsp/issues/1018

tortmayr:
@Sakrafux This change seems to break the `elk-layout-engine` when used in a node context. When executing a layout action the `elk.layout` promise never resolves, hence the operation is never completed and blocks the entire action queue. I guess we have to restrict this changes to the browser-only context.

For now, I'm going to revert this change. Please feel free to open a new PR that fixes this issue.

PR #80, opened 2024-04-24
GLSP-1320: Update to latest glsp dev config

tortmayr:
- Updates to latest version of @eclipse-glsp/dev
- Add generate:index utility script
- Regenerate index files
- Fix codeActionsOnSave in vscode settings
- Refactor upgrade:next script
- Reformat code base with prettier
- Fix copyright headers (by running a full check with glsp checkHeaders . -t full)

Part of https://github.com/eclipse-glsp/glsp/issues/1320

#### Changelog

- [ ] This PR should be mentioned in the changelog


packages/server/src/common/index.ts
```diff
@@ -90,3 +103,4 @@ export * from './utils/layout-util';
 export * from './utils/logger';
 export * from './utils/promise-queue';
 export * from './utils/registry';
+
```
martin-fleck-at, line 106: nit: When I run `yarn generate:index` I only get a single new line in this file. Could you double check by running it on your end?

packages/server/package.json
```diff
@@ -50,6 +50,7 @@
   "scripts": {
     "build": "tsc -b",
     "clean": "rimraf lib *.tsbuildinfo coverage .nyc_output",
+    "generate:index": "glsp generateIndex src/browser src/common src/node -s -f",
```
martin-fleck-at, line 53: Is there a reason we are using `-s` in this repo as opposed to the dedicated index file in each sub-directory?

martin-fleck-at (approved):
Change looks good to me, thank you Tobias! I'm just having two questions I'd like answered before we merge this.

packages/server/package.json
tortmayr: Counter question: Is there a reason for dedicated index files in each subdirectory if we are generating the file anyways?
A single common entry point for the index file seems to be better in this case.
In addition, single index files in each subdirectory might tempt developers (or the TS server autoimport) to import a module from the index instead of its complete path. This can be dangerous since most of our modules are not completly isolated and importing from a (sub) index might lead to nasty circular dependency issues (especially  in combination with inversify).

In general, I would opt for a single index file at the root unless you can think of use case where sub indices are benefical.

martin-fleck-at: I'm fine with the single index file, I was just surprised about the inconsistency between the repositories.

martin-fleck-at (approved):
LGTM!

PR #82, opened 2024-05-15
GLSP-1195: Handle parsing errors in winston logger

tortmayr:
Fixes https://github.com/eclipse-glsp/glsp/issues/1195

#### What it does

- Ensure that the WinstonLogger does not throw an error if it encounters an unparseable object (e.g. circular structure)
- Add fallback hooks for uncaught exceptions/promise rejections to example server to avoid nodes default behavior (process exit)

#### Changelog


martin-fleck-at (approved):
LGTM!

PR #83, opened 2024-05-15
Mocked workerFactory only for webworker context

Sakrafux:
As requested in #78, the change is now restricted to browser/web-worker context.


tortmayr (approved):
Thanks for the rework. LGTM! 👍🏼

PR #87, opened 2024-06-13
Update next dependencies

tortmayr:
#### What it does
- Update next dependencies
- Update changelog
- Update copyright headers
- Add vscode-json rpc as explicit dependency to server package. This ensures that we are still using the same version as the client side in mono repo scenarios where multiple versions of vscode-jsonrpc are resolved

#### Changelog


martin-fleck-at (approved):
LGTM!

PR #88, opened 2024-06-20
Minor cleanup

tortmayr:
- Change visibility of private fields to protected for better extensibility
- Update to latest protocol version
  - This introduces a minor unavoidable API break because GIssue has changed from a class to an interface definition
- Ensure that actions queued with `dispatchAfterNextUpdate` are also dispatched after the initial `SetModelAction`
- Fix copyright header violations

#### Changelog


martin-fleck-at (commented):
Change looks good to me in general but I wonder whether we should have simply deprecated `EdgeSide`, `GIssueSeverity`, and `GIssue` to avoid having a breaking change. What do you think?

tortmayr:
Unfortunately that's not possible because the names clash with the definitions provided by `@eclipse-glsp/protocol`. We cannot export both.
However, in practice the break is really minor, because it only affects direct creation of GIssues on server-side. This is a corner case because typically the server does not directly create issues, instead it creates markers during validation that are applied on the client side and converted to issues.

PR #97, opened 2024-08-28
Fix default type mappings

tortmayr:
#### What it does
Ensure that the correct type for `foreign-object` is configure and add missing configuration for GShaprePreRenderedElements

Fixes https://github.com/eclipse-glsp/glsp/issues/1398

#### Changelog


martin-fleck-at (approved):
LGTM!

PR #99, opened 2024-11-26
Align Node server validation with Java server (async live validation)

martin-fleck-at:
#### What it does

See also https://github.com/eclipse-glsp/glsp-server/pull/247
Fixes https://github.com/eclipse-glsp/glsp/issues/1408

#### How to test

- Add breakpoint during validation and see that the diagram still shows up with the 'Validating model...' message being present.

#### Changelog

### DIFF
--- a/packages/server/package.json
+++ b/packages/server/package.json
@@ -63,11 +63,13 @@
     "@types/uuid": "8.3.1",
     "commander": "^8.3.0",
     "fast-json-patch": "^3.1.0",
+    "lodash": "4.17.21",
     "vscode-jsonrpc": "8.2.0",
     "winston": "^3.3.3",
     "ws": "^8.12.1"
   },
   "devDependencies": {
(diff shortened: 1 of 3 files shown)


tortmayr (commented):
In general the changes look good to me.
However, I'm not the biggest fan of pulling in additional libraries just to use one function.
Especially if its something like debounce which we can easily implement ourself.
I would be in favor of replacing the loadash import with a custom debounce function.

martin-fleck-at:
@tortmayr I agree that pulling in libraries for a single function is not always the best option. However, we already use lodash in the client and in the theia-integration. Do you really think the impact is that large? I'm always a bit worried to re-implement such a common pattern. Note that we can already see the lodash dependency in the `yarn.lock` because we have `concurrently` in our dev (!) dependencies.

tortmayr (approved):

PR #103, opened 2025-02-18
Fix route application

tortmayr:
#### What it does
Fixes https://github.com/eclipse-glsp/glsp/issues/1483
Ensure that the `ComputedBoundsActionHandler` properly applies route i.e. adds source and target point to the corresponding gedge even if the args property of the edge is undefined.

#### Changelog


examples/workflow-server/workspace/example1.wf
planger, line 1: Did you delete that on purpose?

planger (commented):
Looks good otherwise! Thanks!

examples/workflow-server/workspace/example1.wf
tortmayr: I accidentally added this change to this commit. However, yes this was on purpose.
The file is not used anywhere. Having a workspace for the server-only repo does no make sense anyways.
So we can just keep it in^^

planger (approved):
I see, thanks!

PR #105, opened 2025-02-26
Avoid executing unnecessary operations

martin-fleck-at:
#### What it does

- Filter empty change bounds and change routing points operations
- Filter empty compound commands
- Ensure we only submit a new model if actual changes were applied

Relates to  https://github.com/eclipse-glsp/glsp/issues/1477

#### How to test

- Move an element without actually snapping to a new position
- Debug server state or observer undo/redo stack is correctly as there are no operations on the command stack

#### Changelog

-   [x] This PR should be mentioned in the changelog **IF the submission change is accepted as well**
-   [x] This PR introduces a breaking change (if yes, provide more details below for the changelog and the migration guide)
  - If the submission change is accepted as well, this may change the behavior in a way that the GModel is not re-generated after an empty operation. Adopters may have relied on that so this would constitute a breaking change as the previous behavior is also valid and may not strictly be classified as a bug.

### DIFF
--- a/packages/server/src/common/operations/operation-action-handler.ts
+++ b/packages/server/src/common/operations/operation-action-handler.ts
@@ -67,8 +67,9 @@ export class OperationActionHandler implements ActionHandler {
         const command = await handler.execute(operation);
         if (command) {
             await this.executeCommand(command);
+            return this.submitModel();
         }
-        return this.modelSubmissionHandler.submitModel('operation');
+        return [];
     }
 
     protected async executeCommand(command: Command): Promise<void> {
--- a/packages/server/src/common/gmodel/change-bounds-operation-handler.ts
+++ b/packages/server/src/common/gmodel/change-bounds-operation-handler.ts
@@ -13,8 +13,8 @@
  *
  * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
  ********************************************************************************/
-import { GModelRoot, GNode } from '@eclipse-glsp/graph';
-import { ChangeBoundsOperation, Dimension, MaybePromise, Point } from '@eclipse-glsp/protocol';
+import { GModelRoot, GNode, isGBoundsAware } from '@eclipse-glsp/graph';
+import { ChangeBoundsOperation, Dimension, ElementAndBounds, MaybePromise, Point } from '@eclipse-glsp/protocol';
 import { injectable } from 'inversify';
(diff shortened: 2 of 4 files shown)


packages/server/src/common/operations/operation-action-handler.ts
```diff
@@ -67,8 +67,9 @@ export class OperationActionHandler implements ActionHandler {
         const command = await handler.execute(operation);
         if (command) {
             await this.executeCommand(command);
+            return this.submitModel();
         }
-        return this.modelSubmissionHandler.submitModel('operation');
+        return [];
```
martin-fleck-at, line 72: This may change the behavior in a way that the GModel is not re-generated after an empty operation. Adopters may have relied on that so this would constitute a breaking change as the previous behavior is also valid and may not strictly be classified as a bug.

tortmayr: IMO this is more of a bugfix to ensure that the operation handler works as expected.
Nevertheless your are right in certain edge cases adopters may have relied on a model update
(I can`t think of any particular usecase where a effectivly no-op update would be required but who knows).
Maybe we should at least refactor this in overrideable method or property to configure whether update on empty commands should be enabled?
We disable it by default but atleast it gives adopters an easy entry point to change the behavior.

tortmayr (approved):
Changes look good to me.
I guess we need a follow-up for the Java server as well?

packages/server/src/common/operations/operation-action-handler.ts
martin-fleck-at: I agree now that we can consider this more of a bugfix since the [Java server was behaving like this all along](https://github.com/eclipse-glsp/glsp-server/blob/8604c5d1997cf9a743a4fe586f43ff0aa3ee1c0c/plugins/org.eclipse.glsp.server/src/org/eclipse/glsp/server/operations/OperationActionHandler.java#L62-L69). So I would almost argue that we do not need an additional flag. The method as-is is already very short and could be overwritten, don't you think?

martin-fleck-at:
> Changes look good to me. I guess we need a follow-up for the Java server as well?

Follow-up: https://github.com/eclipse-glsp/glsp-server/pull/253

packages/server/src/common/operations/operation-action-handler.ts
tortmayr: 👍🏼 Fine with me.

PR #106, opened 2025-04-03
Ensure ConsoleLogger logs the correct log level of the message

martin-fleck-at:
#### What it does

- Ensure ConsoleLogger logs the correct log level of the message

#### How to test

- See that logged messages using different log levels (info, warn, ...) actually show the correct log level next to the logged message.

#### Changelog

### DIFF
--- a/packages/server/src/common/utils/console-logger.ts
+++ b/packages/server/src/common/utils/console-logger.ts
@@ -16,7 +16,7 @@
 
 import { BindingContext } from '@eclipse-glsp/protocol/lib/di';
 import { injectable } from 'inversify';
-import { LogLevel, Logger, LoggerConfigOptions, LoggerFactory, NullLogger, getRequestParentName } from './logger';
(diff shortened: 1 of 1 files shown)


packages/server/src/common/utils/console-logger.ts
```diff
@@ -33,48 +33,50 @@ export class ConsoleLogger extends Logger {
...
+    protected timestamp(): string {
+        return new Date().toLocaleTimeString();
     }
 
     logAdditionals(...params: any[]): string {
-        if (!params || params.length === 0) {
-            return '';
-        }
-        return params.map(param => this.stringify(param)).join(',\n');
+        const extractedParams = params.length === 1 && Array.isArray(params[0]) ? params[0] : params;
```
tortmayr, line 69: Could you explain the reasoning behind this change?
I don't see why its necessary

tortmayr (approved):

PR #107, opened 2025-04-03
Improve winston configuration

tortmayr:
#### What it does
Avoid unnecessary instanciation of the winston logger if logging is disabled via options (noConsoleLog and noFilelog are true)

#### How to test

Start the server with noConsoleLog enabled. On master this prints a warning since we still instantiate the winston logger although both console and file log are disabled:

```ts
❯ yarn start --no-consoleLog
yarn run v1.22.22
$ yarn --cwd examples/workflow-server-bundled start --no-consoleLog
$ node --enable-source-maps ./wf-glsp-server-node.js --port 5007 --no-consoleLog
[winston] Attempt to write logs with no transports, which can increase memory usage: {"message":"[SocketServerLauncher] The GLSP server is ready to accept new client requests on port: 5007 ","level":"info"}
[GLSP-Server]:Startup completed. Accepting requests on port:5007
```

With the PR this should be fixed.

#### Changelog

### DIFF
--- a/packages/server/src/node/di/app-module.ts
+++ b/packages/server/src/node/di/app-module.ts
@@ -74,6 +74,14 @@ export function configureWinstonLogger<T extends LaunchOptions>(
     rebind = true,
     baseLoggerCreator: (launchOptions: T) => winston.Logger = createWinstonInstance
 ): void {
+    if (rebind) {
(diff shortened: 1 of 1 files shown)


martin-fleck-at (approved):
LGTM!

PR #111, opened 2025-08-25
Use LayoutOperation as optional parameter for layout engine

martin-fleck-at:
#### What it does

Use LayoutOperation as optional parameter for layout engine 

#### Changelog


tortmayr (approved):
LGTM!

PR #120, opened 2025-12-18
Add support for Model Context Protocol

martin-fleck-at:
#### What it does

Exposes each GLSP server as an MCP server over Streamable HTTP, so LLM
clients can query and mutate diagrams via MCP resources, tools, and
prompts.

Architecture
- One MCP endpoint per GLSP server process; under per-window
  deployment, one endpoint per application instance.
- Auto-discovers diagram-scoped contributions from all GLSP client
  sessions and merges them into a shared surface with cross-session
  ID aliasing.
- Pluggable model-serializer, label-provider, and element-types
  extension points for diagram-specific overrides.
- Opt-in: started only when configured on initialize.
- Random port by default; the announced URL is surfaced for adopter
  discovery.
- No built-in auth; loopback-only unless explicitly acknowledged.
- Targets MCP spec 2025-06-18 via the official SDK.

Adopters
- Diagram module: register language-specific tools, prompts,
  resources, and override the diagram extension points.
- Server module: configure server-wide options (port, route, auth
  acknowledgement).
- The workflow server example wires both end-to-end.

Server framework
- Adds an initialize-contribution extension point on the default
  GLSP server, used by MCP to surface configuration on initialize.
  The previous `handleInitializeArgs` hook is deprecated — adopters
  subclassing the default server should migrate.

Documentation
- Adopter README plus a separate architecture document covering the
  feature matrix, lifecycle, extension points, and rationale.

Misc
- Small fix to a base type-guard on edges in the graph package.
- Disposes the server instance when the JSON-RPC client connection
  closes uncleanly (browser/IDE force-close).
- Action dispatcher drops stale response actions arriving without a
  registered handler instead of throwing.

Part of https://github.com/eclipse-glsp/glsp/issues/1546
Requires https://github.com/eclipse-glsp/glsp-client/pull/456

#### How to test

- Start the node server
- Connect with a client that sets the `mcpServer` configuration
- Read the actual url of the server out of the log, e.g., `[McpServerManager] MCP server 'glspMcpServer' is ready to accept new client requests on: http://127.0.0.1:64577/glsp-mcp `
- Connect to the server using the [Theia Integration](https://github.com/eclipse-glsp/glsp-theia-integration/pull/271) (caution: no support for MCP Resources) or [Claude Code](https://code.claude.com/docs/en/mcp) or the [MCP Inspector](https://github.com/modelcontextprotocol/inspector).
- Test querying resources and tools

#### Changelog


packages/server-mcp/src/server/mcp-server-launcher.ts
```diff
@@ -0,0 +1,392 @@
...
+
+    @inject(McpServerOptions) protected mcpOptions: McpServerOptions;
+
+    @inject(McpServerDefaults) protected mcpDefaults: McpServerDefaults;
+
+    @inject(McpHttpTransport) protected transport: McpHttpTransport;
+
+    @inject(GLSPMcpServerFactory) protected glspMcpServerFactory: GLSPMcpServerFactory;
+
+    @inject(McpDiagramHandlerDispatcher) protected dispatcher: DefaultMcpDiagramHandlerDispatcher;
```
tortmayr: Weird mix of symbol injection and direct class typing.

Should probably create an interface for the `McpDiagramHandlerDispatcher` and use it here.

Alternatively, if you dont want to interface this consider renaming to `McpDiagramHandlerDispatcher` and directly using the class as service identifier

packages/server-mcp/src/server/glsp-mcp-server.ts
```diff
@@ -0,0 +1,200 @@
...
+
+    async connect(...args: Parameters<McpServer['connect']>): ReturnType<McpServer['connect']> {
+        await this.mcpServer.connect(...args);
+        // Suppress the expected `RequestTimeout` when no SSE GET stream is open; surface other
+        // failures via debug. `unref` so a forgotten timer can't pin the process.
+        this.keepAliveTimer = setInterval(
+            () =>
+                this.ping().catch(err => {
+                    if (!(err instanceof McpError) || err.code !== ErrorCode.RequestTimeout) {
+                        console.debug('MCP keep-alive ping failed:', err);
```
tortmayr: Consider injecting and using the logger here instead of plain console.log

examples/workflow-server/src/common/graph-extension.ts
```diff
@@ -86,7 +86,7 @@ export class TaskNodeBuilder<T extends TaskNode = TaskNode> extends GNodeBuilder
     protected createCompartmentHeader(): GLabel {
         return new GLabelBuilder(GLabel)
             .type(ModelTypes.LABEL_HEADING)
-            .id(this.proxy.id + '_classname')
+            .id(this.proxy.id + '_label')
```
tortmayr, line 89: Question: Should we update the example1.wf file on the client side to reflect this change?
Or is is purely cosmetically?

packages/server-mcp/src/server/mcp-server-launcher.ts
```diff
@@ -0,0 +1,392 @@
...
+            sessionDisposed: () => this.broadcastResourceListChanged()
+        };
+        this.clientSessionManager.addListener(listener);
+        this.toDispose.push(Disposable.create(() => this.clientSessionManager.removeListener(listener)));
+    }
+
+    /** Best-effort fan-out — failures on individual MCP sessions (e.g. transport mid-close) are swallowed. */
+    protected broadcastResourceListChanged(): void {
+        for (const glspMcpServer of this.sessionServers.values()) {
+            glspMcpServer
```
tortmayr, line 254: SHould we directly expose this on the `GLSPMcpServer` TopLevel ?
Or is this a one time use, and we don't expect adopters to need the same functionality?

examples/workflow-server/src/common/handler/create-merge-node-handler.ts
```diff
@@ -1,5 +1,5 @@
 /********************************************************************************
- * Copyright (c) 2022-2024 STMicroelectronics and others.
+ * Copyright (c) 2022-2026 STMicroelectronics and others.
```
tortmayr: Single header change with out additional file changes, is that correct?

examples/workflow-server/src/common/handler/create-decision-node-handler.ts
```diff
@@ -1,5 +1,5 @@
 /********************************************************************************
- * Copyright (c) 2022-2024 STMicroelectronics and others.
+ * Copyright (c) 2022-2026 STMicroelectronics and others.
```
tortmayr: Single header change with out additional file changes, is that correct?

examples/workflow-server/src/node/app.ts
```diff
@@ -39,15 +40,19 @@ async function launch(argv?: string[]): Promise<void> {
     });
 
     const elkLayoutModule = configureELKLayoutModule({ algorithms: ['layered'], layoutConfigurator: WorkflowLayoutConfigurator });
-    const serverModule = new WorkflowServerModule().configureDiagramModule(new WorkflowDiagramModule(() => GModelStorage), elkLayoutModule);
-
+    const serverModule = new WorkflowServerModule().configureDiagramModule(
+        new WorkflowDiagramModule(() => GModelStorage),
+        elkLayoutModule,
```
tortmayr: Follow-up:
Consider allining this and also provide a ElkLayoutModule base class

packages/graph/src/gedge.ts
```diff
@@ -30,7 +30,7 @@ export class GEdge extends GModelElement {
 
 export namespace GEdge {
```
tortmayr, line 31: Why do we even have this guard? 
We don't have it for any other model classes.
Looks like an remainder of the initial poc implementation ...

Consider deprecating this and using instance of directly

packages/graph/src/gshape-element.ts
```diff
@@ -29,6 +29,12 @@ export abstract class GShapeElement extends GModelElement implements GBoundsAwar
     [GResizable] = true;
 }
 
+export namespace GShapeElement {
```
tortmayr: Im not sure if we really should introduce typeguards for classes.
Simple instance of checks are enough here.

If we do that we probably need to to it consistently for all classes for our GModel...

packages/server-mcp/src/prompts/handlers/describe-diagram-mcp-prompt-handler.spec.ts
```diff
@@ -0,0 +1,83 @@
...
+ * This Source Code may also be made available under the following Secondary
+ * Licenses when the conditions for such availability set forth in the Eclipse
+ * Public License v. 2.0 are satisfied: GNU General Public License, version 2
+ * with the GNU Classpath Exception which is available at
+ * https://www.gnu.org/software/classpath/license.html.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+ ********************************************************************************/
+
+import { ClientSessionManager, Logger, NullLogger } from '@eclipse-glsp/server';
```
tortmayr: Tests for concrete mcp-prompt-handler implements seem unnecessary to me.

You either test basic functionality that should be tested in the the abstract super class
or methods that basically return hardcoded strings

packages/server-mcp/src/prompts/handlers/suggest-improvements-mcp-prompt-handler.spec.ts
```diff
@@ -0,0 +1,60 @@
...
+ * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+ ********************************************************************************/
+
+import { ClientSessionManager, Logger, NullLogger } from '@eclipse-glsp/server';
+import { expect } from 'chai';
+import { Container, ContainerModule } from 'inversify';
+import { McpPromptResult } from '../../server';
+import { SuggestImprovementsArgs, SuggestImprovementsMcpPromptHandler } from './suggest-improvements-mcp-prompt-handler';
+
+/** Build a handler with a {@link ClientSessionManager} stub that returns the supplied session ids. */
```
tortmayr: Tests for concrete mcp-prompt-handler implements seem unnecessary to me.

You either test basic functionality that should be tested in the the abstract super class
or methods that basically return hardcoded strings

packages/server-mcp/src/resources/services/element-types-provider.ts
```diff
@@ -0,0 +1,108 @@
...
+    CreateOperationHandler,
+    Logger,
+    OperationHandlerRegistry
+} from '@eclipse-glsp/server';
+import { inject, injectable } from 'inversify';
+
+/**
+ * A discoverable creatable element type emitted by the `element-types` tool handler.
+ *
+ * Required fields (`id`, `label`) are produced by every provider, including the default
```
tortmayr: I personally don't like the `double` explanation of properties. 
The header JSdoc should be a general summary of the concept/type/module etc.
Property specific documentation should live on property level (and already does so)|
Please clean this up

packages/server-mcp/src/resources/services/element-types-provider.ts
```diff
@@ -0,0 +1,108 @@
...
+ * registry-scrape impl. The optional fields are populated by adopter-specific providers that
+ * have richer semantic info — `description` adds an LLM-facing explanation, `acceptsText`
+ * tells the LLM whether the `text` arg on create-* / modify-* tools is meaningful for this
+ * type. Custom adopters MAY add further passthrough fields beyond these (the schema is
+ * `loose()` on the tool handler's structured output).
+ */
+export interface ElementTypeEntry {
```
tortmayr, line 34: Should  probably also have the @experimental tag.
We need to make sure to properly mark all new introduced types of an experimental api
not only the top level api entry point

packages/server-mcp/src/resources/services/mcp-model-serializer.ts
```diff
@@ -0,0 +1,212 @@
...
+export interface SerializedElement {
+    [key: string]: unknown;
+}
+
+/**
+ * Transforms a graphical model into a representation suitable for LLM consumption. Markdown and
+ * JSON are both reasonable formats; the default impl ({@link MarkdownMcpModelSerializer}) emits
+ * Markdown for {@link serialize}/{@link serializeArray} and an object payload for
+ * {@link serializeStructured} (the dual-emit `structuredContent` counterpart). Aliasing of element
+ * ids happens internally via the per-session {@link McpIdAliasService}; callers don't pass an
```
tortmayr: Drop the ; callers dont pas an alias function

packages/server-mcp/src/server/mcp-tool-handler.spec.ts
```diff
@@ -0,0 +1,182 @@
+/********************************************************************************
```
tortmayr, line 1: Weird naming of test file. Shouldn't this be `mcp-tool-handler.spec.ts`?

packages/server-mcp/src/server/glsp-mcp-server.ts
```diff
@@ -0,0 +1,200 @@
...
+import { injectable } from 'inversify';
+
+export const GLSPMcpServer = Symbol('GLSPMcpServer');
+
+/** Either a static resource or a templated resource registration. */
+export type GLSPMcpResource = RegisteredResource | RegisteredResourceTemplate;
+
+/**
+ * Cadence for server-initiated `ping` requests on the standalone SSE GET stream. Below typical
+ * client-side SSE read timeouts (e.g. undici's 5-min `bodyTimeout`) so the stream stays alive
```
tortmayr: Don't use positional reference to other types in the same file. 
This a) can get stale if the reference is no longer directly "below"
and secondly does not add any value if the comment is viewed standalone (e.g in an IDE preview)

Just remove it

packages/server-mcp/src/server/glsp-mcp-server.ts
```diff
@@ -0,0 +1,200 @@
...
+ * and on overall server shutdown).
+ *
+ * The `GLSP` prefix mirrors core's convention (`GLSPServer`,
+ * `GLSPClientProxy`, `GLSPModule`) where it disambiguates from
+ * external/SDK types. Here it specifically separates `GLSPMcpServer` from
+ * the SDK's `McpServer`; the rest of this package uses an `Mcp` prefix
+ * because there's no name collision.
+ */
+export interface GLSPMcpServer
+    extends Pick<McpServer, 'registerPrompt' | 'registerResource' | 'registerTool' | 'sendLoggingMessage' | 'isConnected' | 'connect'>,
```
tortmayr, line 48: I would prefer explict redeclaration of the picked methods from mcp server here. 
i.e. keep the Pick to ensure compatibility but also redeclare the methods on the interface.

This makes the methods to implement more explicit and allows us to directly add documentation instead 
of convoluting the top-level js doc with property-specific doc.

packages/server-mcp/src/server/glsp-mcp-server.ts
```diff
@@ -0,0 +1,200 @@
...
+    protected readonly resources = new Map<string, GLSPMcpResource>();
+    protected readonly prompts = new Map<string, RegisteredPrompt>();
+    protected keepAliveTimer?: NodeJS.Timeout;
+
+    readonly registerTool: McpServer['registerTool'];
+    readonly registerResource: McpServer['registerResource'];
+    readonly registerPrompt: McpServer['registerPrompt'];
+
+    constructor(
+        protected readonly mcpServer: McpServer,
```
tortmayr: We should probably add the @unmanaged decorators here to be on the safe side.
We had issues in the past with ivnersify and non-injected properties in the constructor without the annotation

packages/server-mcp/src/server/mcp-resource-handler.ts
```diff
@@ -0,0 +1,398 @@
...
+    McpToolResult,
+    extractErrorMessage,
+    requestActionOrFail
+} from './mcp-handler-shared';
+import { McpIdAliasService } from './mcp-id-alias-service';
+import { McpDiagramScopedInput } from './mcp-input-schemas';
+import { McpMimeType } from './mcp-mime-types';
+import { mcpRequestContext } from './mcp-request-context';
+
+/**
```
tortmayr, line 38: Lots of doc file. 
Please double check if everything is necessary 
and trim fluff

packages/server-mcp/src/server/mcp-server-launcher.ts
```diff
@@ -0,0 +1,392 @@
...
+ * This Source Code may also be made available under the following Secondary
+ * Licenses when the conditions for such availability set forth in the Eclipse
+ * Public License v. 2.0 are satisfied: GNU General Public License, version 2
+ * with the GNU Classpath Exception which is available at
+ * https://www.gnu.org/software/classpath/license.html.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+ ********************************************************************************/
+
+import {
```
tortmayr, line 17: Double check comments, trim fluff.

packages/server-mcp/src/server/raw-http.spec.ts
```diff
@@ -0,0 +1,59 @@
...
+ * This Source Code may also be made available under the following Secondary
+ * Licenses when the conditions for such availability set forth in the Eclipse
+ * Public License v. 2.0 are satisfied: GNU General Public License, version 2
+ * with the GNU Classpath Exception which is available at
+ * https://www.gnu.org/software/classpath/license.html.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+ ********************************************************************************/
+
+import * as http from 'http';
```
tortmayr, line 17: Should be in a .spec.ts file. Otherwise it gets shipped as part of the source code

packages/server-mcp/src/tools/handlers/create-nodes-mcp-tool-handler.ts
```diff
@@ -0,0 +1,131 @@
...
+ * This Source Code may also be made available under the following Secondary
+ * Licenses when the conditions for such availability set forth in the Eclipse
+ * Public License v. 2.0 are satisfied: GNU General Public License, version 2
+ * with the GNU Classpath Exception which is available at
+ * https://www.gnu.org/software/classpath/license.html.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+ ********************************************************************************/
+
+import { ApplyLabelEditOperation, CreateNodeOperation } from '@eclipse-glsp/server';
```
tortmayr, line 17: Should this file be tested as well?

packages/server/src/common/operations/create-operation-handler.ts
```diff
@@ -64,3 +64,15 @@ export namespace CreateOperationHandler {
         );
     }
 }
+
+export namespace CreateNodeOperationHandler {
```
tortmayr: Place  the namespace alongside the corresponding interface

packages/server/src/common/protocol/glsp-server.ts
```diff
@@ -103,12 +107,29 @@ export class DefaultGLSPServer implements GLSPServer {
...
         this.initializeResult = result;
         return result;
     }
 
+    protected async initializeServer(params: InitializeParameters, result: InitializeResult): Promise<InitializeResult> {
+        for (const initializer of this.initializers) {
+            try {
+                result = await initializer.initializeServer(this, params, result);
+            } catch (error: unknown) {
+                this.logger.error(`Error during server initialization from ${initializer.constructor.name}:`, error);
```
tortmayr, line 124: Catch without rethrow breaks promise resolution. 
We catch all errors -> promise returned is never rejected

tortmayr (changes requested):
Thanks for the rework Martin. 
In general the code looks good to  me. 
I have a couple of inline comments.

In addition, we should probably create follow-up issues for the following tasks

- Java alignment/implementation. Not planned in yet from our side, but we should open the issue nevertheless for tracking. (tag with help wanted, looking for sponsor)

- Once this is merged we should update the protocol doc on the website

- It seems like we have a couple of concepts in place (initializers, factories) which sole purpose is to bypass inversify circular dependencies for contribution points. We should consider an alignmnet here and use the `LazyInjector` approach consistently (same as in client)

packages/graph/src/gedge.ts
martin-fleck-at: Yeah, I was wondering about that too. I'm happy to deprecate it, thanks!

examples/workflow-server/src/node/app.ts
martin-fleck-at: Good idea!

martin-fleck-at:
> Thanks for the rework Martin. In general the code looks good to me. I have a couple of inline comments.

Thank you very much, I'll address them shortly!

> * Java alignment/implementation. Not planned in yet from our side, but we should open the issue nevertheless for tracking. (tag with help wanted, looking for sponsor)
cf https://github.com/eclipse-glsp/glsp/issues/1672

> * Once this is merged we should update the protocol doc on the website
Good point! I added https://github.com/eclipse-glsp/glsp/issues/1673 to track this officially.

> * It seems like we have a couple of concepts in place (initializers, factories) which sole purpose is to bypass inversify circular dependencies for contribution points. We should consider an alignmnet here and use the `LazyInjector` approach consistently (same as in client)

Do you have any particular places in mind that we introduced? I think we only have the registry initializers, similar to the `ActionHandlerRegistryInitializer` and `OperationHandlerRegistryInitializer` that already exist. And then we have a few factories similar to `OperationHandlerFactory` and `ActionHandlerFactory`. Or is that a more general comment to align long-term?

packages/server-mcp/src/server/mcp-server-launcher.ts
martin-fleck-at: I wouldn't expect adopters to overwrite this tbh. we are the only caller and then we would also need to expose other similar methods. If you agree, I think wee should leave it as is.

tortmayr:
> Do you have any particular places in mind that we introduced? I think we only have the registry initializers, similar to the `ActionHandlerRegistryInitializer` and `OperationHandlerRegistryInitializer` that already exist. And then we have a few factories similar to `OperationHandlerFactory` and `ActionHandlerFactory`. Or is that a more general comment to align long-term?

That was more a general comment to align long term. These concepts were introduced before the `LazyInjector` API was available. Now, it think it makes sense to align this and use one consistent approach for initializing contributions.

packages/graph/src/gshape-element.ts
martin-fleck-at: Yeah, let's not, I'll remove it again.

examples/workflow-server/src/common/graph-extension.ts
martin-fleck-at: The file is already correct, just the server was inconsistent. We can leave it I believe.

martin-fleck-at:
I pushed a commit that hopefully addresses all your concerns.

tortmayr (approved):
LGTM! Thanks for the fast update.
👍🏼

PR #127, opened 2026-03-25
Issues/1546

Sakrafux:
Extends the GLSP MCP Server started in [PR-120](https://github.com/eclipse-glsp/glsp-server-node/pull/120) according to our meetings @martin-fleck-at.


packages/server-mcp/src/server/mcp-option-service.ts
```diff
@@ -0,0 +1,53 @@
...
+
+export const McpOptionService = Symbol('McpOptionService');
+
+export interface McpOptionService {
+    set(options: McpServerOptions): void;
+    get(key: keyof McpServerOptions): boolean | undefined;
+}
+
+@injectable()
+export class DefaultMcpOptionService implements McpOptionService {
```
martin-fleck-at: This seems a bit over-engineered just to expose the options map since it is only set once and then only read.

packages/server-mcp/src/tools/handlers/delete-elements-handler.ts
```diff
@@ -0,0 +1,104 @@
...
+        // Validate elements exist
+        const missingIds: string[] = [];
+        const realIds: string[] = [];
+        for (const elementId of elementIds) {
+            const realId = mcpIdAliasService.lookup(sessionId, elementId);
+            const element = modelState.index.find(realId);
+            if (element) {
+                realIds.push(realId);
+            } else {
+                missingIds.push(elementId);
```
martin-fleck-at: Should we push the `elementId` or `realId` here? We expose it as part of the error message so we need to make sure that we consistently use either the alias IDs or the real IDs, depending on who is supposed to deal with the message (human vs LLM).

packages/server-mcp/src/tools/handlers/get-selection-handler.ts
```diff
@@ -0,0 +1,109 @@
...
+export class GetSelectionMcpToolHandler implements McpToolHandler, ActionHandler {
+    actionKinds = [GetSelectionMcpResultAction.KIND];
+
+    @inject(Logger)
+    protected logger: Logger;
+
+    @inject(ClientSessionManager)
+    protected clientSessionManager: ClientSessionManager;
+
+    protected resolvers: Record<string, { sessionId: string; resolve: (value: CallToolResult | PromiseLike<CallToolResult>) => void }> = {};
```
martin-fleck-at: I'm not a big fan of this but until https://github.com/eclipse-glsp/glsp/issues/607 is resolved, there really is not good alternative. We should definitely tackle that next.

martin-fleck-at (commented):
Hi @Sakrafux! Thank you for the great overall work! I do have some comments and questions regarding parts of the code but I am a big fan of the overall ideas and configurability. Since we are not merging into main (yet), this definitely does not need to be perfect but I would love to get your thoughts and feedback on some of the comments.

packages/server-mcp/src/server/mcp-option-service.ts
Sakrafux: Yeah, I know, but the alternative is heavy prop drilling.

For context, I need a way for the `McpIdAliasService` to know, whether the feature is active. I can't register 2 different versions of this service, because I only know of the feature flag at runtime when the server is initialized. Therefore, I need to get this information into the `alias` and `lookup` methods of that service. I could add it to the parameters, but as this service is heavily used in `McpModelSerializer`, I would also need to add it there as well, which I really don't like. The alternative is to just write it into the inversify container context via this helper service.

However, I am not that well versed in inversify, so there may very well be a better way to communicate that information at runtime.

packages/server-mcp/src/server/mcp-server-manager.ts
```diff
@@ -52,8 +69,11 @@ export class McpServerManager implements GLSPServerInitContribution, GLSPServerL
             return result;
         }
 
-        const { port = 0, host = '127.0.0.1', route = '/glsp-mcp', name = 'glspMcpServer' } = mcpServerParam;
-        const mcpServerConfig: FullMcpServerConfiguration = { port, host, route, name };
+        // use a fixed default port instead of 0 so that the MCP server need only be registered once
+        // using 0, i.e., a random port, would require re-registering the MCP server each time
+        const { port = 60000, host = '127.0.0.1', route = '/glsp-mcp', name = 'glspMcpServer', options = {} } = mcpServerParam;
```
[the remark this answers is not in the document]
Sakrafux: The problem with using `port = 0` is that each startup creates a wholly new MCP server to register. Assuming that most users don't actually start multiple GLSP servers (and thus MCP servers) at once (e.g., Theia or VSCode only start a single one regardless of the number of tabs), I would say its better to use an actual default port. If it becomes an issue, just setting different params is sufficient.

packages/server-mcp/src/resources/services/mcp-model-serializer.ts
```diff
@@ -0,0 +1,181 @@
...
+ * for communicating with an LLM. It is recommended to use Markdown or simply JSON for this purpose.
+ */
+export interface McpModelSerializer {
+    /**
+     * Transforms the given {@link GModelElement} into a string representation.
+     * @param element The element that should be serialized.
+     * @param aliasFn Optional function to alias an ID according to `McpIdAliasService`
+     * @returns The transformed string and the underlying flattened graph object.
+     */
+    serialize(element: GModelElement, aliasFn?: (id: string) => string): [string, Record<string, Record<string, any>[]>];
```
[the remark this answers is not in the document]
martin-fleck-at: I see. I think I would rather remove the return type from the API since - as you said yourself - it is more of an implementation detail (we do not need it for Markdown but we might want it form JSON) and the GLSP infrastructure does not depend on it. If someone needs it, they can always provide their own serializer implementation and inject that wherever they need the specifics of that particular serializer.

PR #131, opened 2026-04-09
feat: add request/response support to server ActionDispatcher

martin-fleck-at:
#### What it does

Enable the server to send requests to the client and await responses, and to issue requests handled locally by server-side handlers. Complements the client-side changes in glsp-client.

- Add `request()` and `requestUntil()` to ActionDispatcher
- Intercept responses in `dispatch()` and `doDispatch()` to resolve pending requests
- Translate `RejectAction` responses to promise rejections
- Bypass action queue for nested requests to prevent deadlocks
- Tighten `shouldForwardToClient()` to use `hasValidResponseId()`
- Add async `handleClientRequest()` in `DefaultGLSPServer.process()`
- Add tests for request/response, deadlocks, timeouts, late responses, and dispose cleanup

Relates to https://github.com/eclipse-glsp/glsp/issues/607

#### How to test

- Try together with https://github.com/eclipse-glsp/glsp-client/pull/480
- Add some request actions (selection, editor context, etc.) to a server-side handler and see that you receive the result as expected.

#### Changelog

-   [x] This PR introduces a breaking change (if yes, provide more details below for the changelog and the migration guide)

It adds two methods to the server-side action dispatcher. We do have default implementations but if adopters have their own implementation, they will be missing those methods.

**Note:** Will break until the protocol changes in https://github.com/eclipse-glsp/glsp-client/pull/480 are merged.


packages/server/src/common/actions/action-dispatcher.ts
```diff
@@ -65,6 +67,45 @@ export interface ActionDispatcher {
...
     dispatchAfterNextUpdate(actions: Action[]): void;
     dispatchAfterNextUpdate(...actions: Action[]): void;
+
+    /**
+     * Dispatches a request action and returns a promise that resolves when a matching response
+     * action is dispatched or rejects if the response is a {@link RejectAction}. The response is
+     * _not_ passed to the registered action handlers. Instead, it is the responsibility of the
+     * caller of this method to handle the response properly.
+     *
+     * If the request's `kind` is registered in `ClientActionKinds`, it is forwarded to the client
```
tortmayr, line 79: IMO I think this is the part that needs some testing and hardening.

For the request-response feature we work on the implicit assumption that the handling side has exactly one action handler that returns the corresponding response.  Multiple handlers for the action kind can be registered, but they either need to be void or return other action/command types.

We currently have no way to verify that assumption. So what happens if we have two registered response handlers for a request.
- Which one wins?
- What happens to the response action of the second handler? Is it rejected, dispatched normally, erroneous at the receiving side etc.

Similar to that, what happens if both the client and server have registered a response handler for the request action.
This is implicit a flawed design since we have this `only one response handler` restriction.
But are these cases handled gracefully?
Which side wins. does this cause errors on the other side, is there logging/usable output for the adopter if the run into this case etc.

Maybe we should continue the discussion to this topic in a different channel (i.e. face-to-face meeting)

packages/server/src/common/actions/action-dispatcher.ts
```diff
@@ -83,8 +124,26 @@ export class DefaultActionDispatcher implements ActionDispatcher, Disposable {
...
     dispatch(action: Action): Promise<void> {
+        // Fast-path: resolve pending requests before the queue to prevent deadlock
+        // when request() is awaited inside a queued handler and the response
+        // arrives via process() -> dispatch().
+        if (this.interceptPendingResponse(action)) {
+            return Promise.resolve();
+        }
+
         // Dont queue actions that are just delegated to the client
         if (this.clientActionForwarder.shouldForwardToClient(action)) {
```
tortmayr: Not directly related to this PR, but maybe an issue that we should track in  a ticket:

This queue skipping for client actions could be problematic if the server has also registered an handler for this action kind.

In that case, the server handled actions escapes the sequential order is dispatched directly before any other queued server actions.

packages/server/src/common/actions/action-dispatcher.ts
```diff
@@ -144,8 +207,149 @@ export class DefaultActionDispatcher implements ActionDispatcher, Disposable {
...
+            }
+            // Drain post-update actions before resolving — in the intercept path
+            // there's no responses array, so dispatch them directly.
+            const postUpdateActions = this.drainPostUpdateQueue(action);
+            if (RejectAction.is(action)) {
+                deferred.reject(new Error(`${action.message}${action.detail ? ': ' + action.detail : ''}`));
+            } else {
+                deferred.resolve(action);
+            }
+            if (postUpdateActions.length > 0) {
```
tortmayr, line 386: Shouldn't we handle the case of rejeced `SetModel` `UpdateModel` action here with an explicit error.
i.e. if the action was rejected but we have postUpdateModel actions we are in a potentially fatal state.
The core model update went wrong and we should inform the user.

packages/server/src/common/protocol/glsp-server.ts
```diff
@@ -150,9 +153,37 @@ export class DefaultGLSPServer implements GLSPServer {
...
+        clientSession: ClientSession,
+        action: RequestAction<ResponseAction>,
+        clientId: string
+    ): Promise<void> {
+        try {
+            const response = action.timeout !== undefined
+                ? await clientSession.actionDispatcher.requestUntil(action, action.timeout, true)
+                : await clientSession.actionDispatcher.request(action);
+            if (response) {
+                this.sendToClient({ clientId, action: response });
```
tortmayr: Similar to client side. Maybe extract this into submethod for more user control.

tortmayr (changes requested):
Here are my review comments after a first high level review and quick testing.

One additional issue:

We currently have no actual server side request in the example code i.e. no way to test the feature e2e or manually. We currently solely rely on unit tests.
Maybe it would be a good to exetend the workflow example with  handler that actually uses the new request handling. Could be something simlple like retrieving the current editor context from the client and logging it.

packages/server/src/common/actions/action-dispatcher.ts
martin-fleck-at: I adapted the code now that client actions do not skip the queue automatically but instead we have a proper `dispatchDirectly` to dispatch actions without the queue, e.g., for progress reporting.

packages/server/src/common/actions/action-dispatcher.ts
martin-fleck-at: If we run into a project, we should have a RejectAction anyway and the post update actions are empty.

packages/server/src/common/protocol/glsp-server.ts
martin-fleck-at: Good idea!

packages/server/src/common/actions/action-dispatcher.ts
martin-fleck-at: I documented this assumption of only the first response handler resolving the request, any other responses (late, second responses or responses without request) are simply dispatched as normal actions, as they are on the client.

martin-fleck-at:
@tortmayr I updated the PR and now we have a more explicit handling of actions:
- `dispatch`: Dispatches actions in order, i.e., queues them. That is true for server actions and client actions.
- `dispatchDirectly`: Dispatches actions immediately without queue. That is true for handler return actions or if called explicitly (e.g., progress reporting).
- `interceptPendingResponse` (internal): Responses to requests do not need to be re-dispatched but simply resolve the request. If there is no matching request, we queue the action.

As for the example, there really is no good place in the workflow example to do something semantically useful. However, with the upcoming MCP feature, we will need it for the tool functions and there the usage will be well-tested. Is that enough?

packages/server/src/common/actions/action-dispatcher.ts
```diff
@@ -56,6 +60,18 @@ export interface ActionDispatcher {
...
+     * Dispatches an action directly, bypassing the action queue. Use this for actions that
+     * need to be processed immediately, e.g. progress notifications sent from inside a handler.
+     *
+     * Actions dispatched this way are not sequenced with other queued actions. Callers are
+     * responsible for ensuring that concurrent execution is safe.
+     *
+     * @param action The action to dispatch directly.
+     * @returns A promise indicating when the action processing is complete.
+     */
+    dispatchDirectly(action: Action): Promise<void>;
```
tortmayr: As discussed offline, I think this change is a significant behavioral break for adopters. Instead we should use the same approach as in the Java Server. i.e. all actions that are dispatched or returned from within an action handler are bypassing the queue and dispatched directly.

We can use an `AsyncLocalStorage` (node builtin) to detect these actions.

tortmayr (changes requested):
Following our discussion offline, I would like to see the following changes

- Revert the introduction of dispatchDirectly. Instead use the same approach as
 in the Java Server. (Actions that are dispatched/returned from within an action handler bypass the queue). `AsyncLocalStorage` can be used for this
- Replace custom `PromiseQueue`. Our current custom promise queue is a dated custom implementation that has a couple of issues in particular with error delegation/swalloing and unnecessary overheads.The current `PromiseQueue` should be replaced with a channel + consumer loop pattern that mirrors the Java architecture directly.
The pattern could look like this:

```ts
class ActionChannel<T> {
    private queue: Array<{
        item: T;
        resolve: (v: void) => void;
        reject: (e: any) => void;
    }> = [];
    private notify: (() => void) | null = null;
    private stopped = false;

    push(item: T): Promise<void> {
        if (this.stopped) {
            return Promise.reject(new Error('Channel is closed'));
        }
        return new Promise((resolve, reject) => {
            this.queue.push({ item, resolve, reject });
            this.notify?.();
        });
    }

    async *consume(): AsyncGenerator<{
        item: T;
        resolve: (v: void) => void;
        reject: (e: any) => void;
    }> {
        while (!this.stopped || this.queue.length > 0) {
            while (this.queue.length > 0) {
                yield this.queue.shift()!;
            }
            if (!this.stopped) {
                await new Promise<void>(r => { this.notify = r; });
            }
        }
    }

    stop(): void {
        this.stopped = true;
        this.notify?.();
    }

    rejectPending(): void {
        for (const entry of this.queue) {
            entry.reject(new Error('Channel cleared'));
        }
        this.queue = [];
    }

    get size(): number { return this.queue.length; }
}
````

In addition, we should get rid of the second ephemeral queue we use for ordering responses. It is completely unnecessary instead we can just use a for-each await loop.

File: packages/server/src/common/actions/action-dispatcher.ts (hunk 5 above)
Comment (martin-fleck-at) [reply packages/server/src/common/actions/action-dispatcher.ts]: Good idea, will do!

**Comment by martin-fleck-at** [comment]
> * Revert the introduction of dispatchDirectly. Instead use the same approach as
>   in the Java Server. (Actions that are dispatched/returned from within an action handler bypass the queue). `AsyncLocalStorage` can be used for this

Done, the action dispatcher is now much more aligned with the Java behavior! I was not aware of AsyncLocalStorage but that is pretty cool!

> * Replace custom `PromiseQueue`. Our current custom promise queue is a dated custom implementation that has a couple of issues in particular with error delegation/swalloing and unnecessary overheads.The current `PromiseQueue` should be replaced with a channel + consumer loop pattern that mirrors the Java architecture directly.

I replaced the existing PromiseQueue and marked it deprecated.

File: packages/server/src/common/actions/action-dispatcher.ts [hunk 6]
```diff
@@ -81,15 +135,55 @@ export class DefaultActionDispatcher implements ActionDispatcher, Disposable {
     @inject(ClientId)
     protected clientId: string;
 
-    protected actionQueue = new PromiseQueue();
+    protected channel = new ActionChannel<Action>();
+    protected dispatchContext = new AsyncLocalStorage<boolean>();
+    protected consumerLoop: Promise<void> | undefined;
```
tortmayr: What is the point of storing the consumer loop promise as a property?
It is currently unused. Was the intend to dispose the promise?
If so its missing 😉

packages/server/src/common/actions/action-dispatcher.ts
```diff
@@ -42,6 +45,15 @@ export interface ActionDispatcher {
     /**
```
tortmayr, line 44: Newly added comments in the ActionDispatcher interface leak implementation details of the the default implementation (Channel, localStorage etc.) Should probably be more concise and generic

packages/server/src/common/actions/action-dispatcher.ts
```diff
@@ -81,15 +135,55 @@ export class DefaultActionDispatcher implements ActionDispatcher, Disposable {
...
+    protected readonly timeouts = new Map<string, NodeJS.Timeout>();
+    protected nextRequestId = 1;
+
+    @postConstruct()
+    protected initialize(): void {
+        this.consumerLoop = this.runConsumerLoop();
+    }
+
+    /**
+     * Generates a unique request ID for this dispatcher. The `clientId` prefix makes IDs
```
tortmayr: Method is self explanatory. Does it really warrant a comment?

packages/server/src/common/actions/action-dispatcher.ts
```diff
@@ -81,15 +135,55 @@ export class DefaultActionDispatcher implements ActionDispatcher, Disposable {
...
+     */
+    protected generateRequestId(): string {
+        return `server_${this.clientId}_${this.nextRequestId++}`;
+    }
 
     dispatch(action: Action): Promise<void> {
-        // Dont queue actions that are just delegated to the client
-        if (this.clientActionForwarder.shouldForwardToClient(action)) {
+        // Order matters:
+        // 1. Responses to pending request() calls must short-circuit here; queueing would
```
tortmayr: If you want to keep that comment please make it more concise.
I especially don see point in loosely referencing java implementations

packages/server/src/common/actions/action-dispatcher.ts
```diff
@@ -144,8 +236,152 @@ export class DefaultActionDispatcher implements ActionDispatcher, Disposable {
...
+            clearTimeout(staleTimeout);
+            this.timeouts.delete(action.responseId);
+            this.logger.debug(`Late response for timed-out request '${action.responseId}', dispatching as normal action`);
+            action.responseId = '';
+        }
+        return false;
+    }
+
+    /**
+     * Shuts down the dispatcher. Stops the channel, rejects queued actions and pending request()
```
tortmayr: I don't thinks is necessary to comment all details of a simple dispose method 😉

packages/server/src/common/utils/action-channel.ts
```diff
@@ -0,0 +1,101 @@
...
+ */
+export interface ActionChannelEntry<T> {
+    item: T;
+    resolve: () => void;
+    reject: (error: unknown) => void;
+}
+
+/**
+ * Producer/consumer channel with a single async consumer loop. Mirrors the Java
+ * dispatcher's `BlockingQueue` + consumer thread architecture.
```
tortmayr: As I mentioned before I'm not sure if we should loosely reference implementation details of the Java server.
The benefit is not really high and the potential for stale doc is rather big.

packages/server/src/common/utils/action-channel.ts
```diff
@@ -0,0 +1,101 @@
...
+            this.notify?.();
+        });
+    }
+
+    /**
+     * Yields pending entries in FIFO order, blocking on a notification promise when empty.
+     * Exits once the channel is stopped and the queue has been drained.
+     */
+    async *consume(): AsyncGenerator<ActionChannelEntry<T>> {
+        while (!this.stopped || this.queue.length > 0) {
```
tortmayr: What happens if `ActionChannel.consume()` is called multiple.
I assume it would  yield items to both consumers non-deterministically.
In praxis this currently does never happen, but we should probably still guard it somehow.

tortmayr (commented):
In general, changes look good to me.

We now have one major issue.  async-hooks are a  node-specfic and cannot be used in the browser context.
We need to move out the node specifics of the action dispatcher from the common package and also 
think about how we implement this for the browser-only side since `AsyncLocalStorage` is not available there.

tortmayr:
To follow up on the browser issue:

  Proposed fix: Inject `AsyncLocalStorage` via DI instead of importing it directly.

  - Define a DI token in `src/common/di/service-identifiers.ts` (e.g. AsyncLocalStorageProvider)
  - src/node/ binds the native `AsyncLocalStorage` from async_hooks
  - src/browser/ binds `AsyncLocalStorage` from https://www.npmjs.com/package/als-browser — a zero-dependency, drop-in polyfill b
  by Node.js core contributor Stephen Belanger (used transitively by dc-browser at ~245k weekly downloads)

  This keeps action-dispatcher.ts in src/common/ with no code duplication.

packages/server/src/common/actions/action-dispatcher.ts
martin-fleck-at: Ah, yes, exactly! I was intending to do a proper disposal but since we cannot have async drain-dispose anyways, I forgot to remove it again. Good catch!!

packages/server/src/common/actions/action-dispatcher.ts
martin-fleck-at: No, removed it.

packages/server/src/common/actions/action-dispatcher.ts
martin-fleck-at: I shortened it a bit but I do want to make sure we document that the order is important and why some actions go one way instead of the other.

packages/server/src/common/utils/action-channel.ts
martin-fleck-at: You are right, I'll throw an error to indicate that this should not happen.

martin-fleck-at:
@tortmayr I pushed an update that should fix the issues you mentioned! Great idea with the browser-variant of the AsyncLocalStorage!

martin-fleck-at:
CI is failing because we need to merge https://github.com/eclipse-glsp/glsp-client/pull/480 first.

tortmayr:
@martin-fleck-at Thanks for the update. During review I noticed that we currently have no browser-based workflow example to test if the dispatching changes work in the browser flow.
We should create a follow-up for that.
In addition, we also need a follow-up to apply the request-response changes to the java-server

I created a browser-based workflow example locally for testing, which I can contribute once cleaned up.

During testing I noticed, that unfortunately the used ALS polyfill does not work reliably in all browser/js engines.
In particular it breaks for chromes V8 engine. When used in chrome direct response actions of an handler are treated as out-of-context i.e. they are queued and create a deadloop.

Unfortunately, there are not real ALS alternatives available so we have to use a custom solution.
I pushed a commit with a simple flag based solution for the browser, that gives us more or less the same behavior for the ActionDispatcher as ALS would.
There is one minor corner case where the current solution would work differently (it is noted in the documentation of `BrowserDispatchContext`. However I think its worth the tradeoff.

We can discuss this in more detail face-to-face.

WDYT?

martin-fleck-at:
@tortmayr Overall the code looked good to me. I still removed the als-browser dependency and cleaned up some other things, especially in naming things. Do you want to have another look?

packages/server/src/common/actions/action-dispatcher.ts
```diff
@@ -161,19 +167,21 @@ export class DefaultActionDispatcher implements ActionDispatcher, Disposable {
     @inject(ClientId)
     protected clientId: string;
 
-    @inject(ActionDispatchContext)
-    protected dispatchContext: ActionDispatchContext;
+    @inject(ActionDispatchScope)
+    protected dispatchScope: ActionDispatchScope;
 
-    protected channel = new ActionChannel<Action>();
+    protected actionQueue = new ActionChannel<Action>();
```
tortmayr: What about the class name itself should we also rename this to `ActionQueue`?

packages/server/src/common/actions/action-dispatcher.ts
```diff
@@ -350,15 +366,15 @@ export class DefaultActionDispatcher implements ActionDispatcher, Disposable {
         if (!ResponseAction.hasValidResponseId(action)) {
             return false;
         }
-        // Node-only: responses to server-initiated requests are resolved here instead of going
-        // through action handlers. No Java equivalent.
+        // Responses to server-initiated requests are resolved here instead of going through
```
tortmayr: Is the Java-related  statement even true? We don't know yet right since the java-implementation is a follow-up.
We should probably remove it

packages/server/src/node/di/app-module.ts
```diff
@@ -16,15 +16,17 @@
...
 import { LaunchOptions } from '../launch/cli-parser';
-import { NodeDispatchContext } from './node-dispatch-context';
+import { NodeActionDispatchScope } from './node-action-dispatch-scope';
 import { WinstonLogger } from './winston-logger';
 
 export function createAppModule(options: LaunchOptions): ContainerModule {
     return new ContainerModule((bind, unbind, isBound, rebind) => {
         bind(InjectionContainer).toDynamicValue(dynamicContext => dynamicContext.container);
-        bind(ActionDispatchContext).toDynamicValue(() => new NodeDispatchContext());
+        // Transient on purpose: kept symmetric with the browser binding, which cannot share a
```
tortmayr: Is symmetry here really important?
If we can use one singleton implementation on the node side thanks to ALS why shouldn't we do it?

tortmayr (changes requested):
Thanks. I have a few minor inline comments, but other than that LGTM.

packages/server/src/common/actions/action-dispatcher.ts
martin-fleck-at: You are right, we do plan to tackle that shortly, so I'll remove it.

packages/server/src/common/actions/action-dispatcher.ts
martin-fleck-at: Yes, you are right, let's do that! good catch!!

packages/server/src/node/di/app-module.ts
martin-fleck-at: I'm fine either way, it does not really make a difference and I thought for re-binding it might be simpler but I'll change it.

martin-fleck-at:
@tortmayr I think all points should be addressed in the new commit.

tortmayr (approved):
LGTM! 👍🏼

PR #132, opened 2026-04-29
GLSP-1565: Publish browser workflow-server bundle

tortmayr:
#### What it does

Add new `@eclipse-glsp-examples/workflow-server-bundled-web` package that publishes the pre-built webworker bundle with inline source maps.

- Create `examples/workflow-server-bundled-web/` package
- Update webpack config to output webworker bundle to the new package
- Use inline source maps for the webworker target
- Add browser bundle to root `bundle` script
- Remove redundant "Bundle browser example" CI step

Part of: https://github.com/eclipse-glsp/glsp/issues/1656

#### Changelog

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,9 +22,6 @@ jobs:
           node-version: 22.x
       - name: Build
         run: yarn install
-      - name: Bundle browser example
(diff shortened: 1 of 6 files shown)


martin-fleck-at (approved):
LGTM!

PR #133, opened 2026-04-30
GLSP-1565: Use separate source maps for web bundle

tortmayr:
## Summary
Follow-up to #132.

- Switch from inline to separate source map files for the webworker bundle
- Include `.map` file in published package
-  Sync example1.wf with client variant

## Test plan
- [ ] `yarn build` produces both `.js` and `.js.map` in `examples/workflow-server-bundled-web/`


martin-fleck-at (approved):
LGTM!

PR #134, opened 2026-04-30
fix: correct publish workflow branch filter

tortmayr:
## Summary
- The `workflow_run` trigger in `publish.yml` was filtering on `master`, but the default branch is `main`
- This caused auto-publishing to never fire after CI completed on `main`


martin-fleck-at (approved):
;-)

PR #135, opened 2026-05-13
Generate and use only single index per repository convention

martin-fleck-at:
#### What it does

Generate and use only single index per repository convention

#### Changelog


tortmayr (approved):
Looks good. 
One question, do we also need to change the script in the package.json for index generation?

martin-fleck-at:
No, actually I think we are fine with what we have.

PR #136, opened 2026-05-18
Make MCP server browser-portable with a Web-standard handler

martin-fleck-at:
#### What it does

Replace the Hono-only Node handler with a Fetch-API handler that runs on Node, Bun, Deno, Cloudflare Workers, and in-page Web Workers. Restructure @eclipse-glsp/server-mcp into a browser/common/node split mirroring @eclipse-glsp/server.

- New browser entry point with a Web Worker integration helper.
- VS Code web-extension bundle no longer transitively pulls Node-only deps.
- Mutating tools report dispatched-command counts for cross-call undo / redo.
- Port-agnostic Host-header validation — matches the SDK's older Express middleware that the Fetch transport regressed.
- Workflow example ships an end-to-end browser demo of the portable handler.
- Non-loopback auth and shared session state remain adopter responsibility.
- ARCHITECTURE.md refreshed for the portable handler and for the IDE integrations that now register the GLSP MCP server with the host IDE.

Part of https://github.com/eclipse-glsp/glsp/issues/1546

#### How to test

- Everything should work as before in our classic cases.
- Try the browser entry in the bundled browser server test app.

#### Changelog


examples/workflow-server/src/browser/app.ts
```diff
@@ -17,12 +17,19 @@ import 'reflect-metadata';
...
 import { Container } from 'inversify';
 import { WorkflowLayoutConfigurator } from '../common/layout/workflow-layout-configurator';
+import { WorkflowMcpDiagramModule } from '../common/mcp/workflow-mcp-diagram-module';
 import { WorkflowDiagramModule, WorkflowServerModule } from '../common/workflow-diagram-module';
 import { WorkflowMockModelStorage } from './mock-model-storage';
 
-export async function launch(argv?: string[]): Promise<void> {
+export async function launch(_argv?: string[]): Promise<void> {
+    // Bridge must be created before any await so postMessages that arrive on the next event-loop
+    // tick aren't dropped. Requests pile up on the bridge's internal `launcherReady` until the
```
tortmayr: Keep concise. First sentence is enough

examples/workflow-server-mcp-demo/README.md
```diff
@@ -0,0 +1,67 @@
...
+`packages/server-mcp/src/node/server/mcp-http-transport-e2e.spec.ts`, which runs an MCP SDK
+`Client` over real HTTP against the launcher.
+
+## Running
+
+From the repository root:
+
+```bash
+yarn workspace @eclipse-glsp-examples/workflow-server build
+yarn workspace @eclipse-glsp-examples/workflow-server-mcp-demo start
```
tortmayr: `yarn start:mcp-demo` should be fine here right?

packages/server-mcp/src/browser/server/browser-mcp-request-context.ts
```diff
@@ -0,0 +1,50 @@
...
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+ ********************************************************************************/
+
+import { injectable } from 'inversify';
+import { McpRequestContext, McpRequestExtra } from '../../common/server/mcp-request-context';
+
+/**
+ * Browser-compatible {@link McpRequestContext}. Single mutable slot — concurrent requests on
+ * the same session overwrite each other's frame. Hosts must serialise handler chains.
```
tortmayr: The comment says "hosts must serialise
handler chains" and the McpWorkerBridge does this via dispatchChain. This is correct but fragile — if an adopter uses
BrowserMcpServerModule directly without the bridge's serialization, concurrent requests will silently corrupt the context.
Consider logging a warning or documenting this more prominently.

packages/server-mcp/src/browser/server/mcp-worker-bridge.ts
```diff
@@ -0,0 +1,172 @@
...
+            response.headers.forEach((value, key) => {
+                headers[key] = value;
+            });
+            this.reply(port, {
+                type: MCP_RESPONSE_MESSAGE_TYPE,
+                id: request.id,
+                status: response.status,
+                statusText: response.statusText,
+                headers,
+                body: response.body
```
tortmayr: The type says ReadableStream<Uint8Array> | null
  and it's passed as a Transferable in reply(). However, ReadableStream is only transferable in Chromium (Firefox and Safari
  don't support it). The README already says "Chromium-based browser" but this limits portability.

tortmayr (changes requested):
Changes look good to me. 
Clean split into browser/node entrypoints.
And the MCP browser demo is very nice 👍🏼

examples/workflow-server/src/browser/app.ts
martin-fleck-at: Good call, trimmed to the first sentence.

examples/workflow-server-mcp-demo/README.md
martin-fleck-at: Yep, switched to yarn start:mcp-demo. Thanks!

packages/server-mcp/src/browser/server/browser-mcp-request-context.ts
martin-fleck-at: Good point. Added a one-shot console.warn when run() is entered while another context is still active, and expanded the doc comment to point adopters at McpWorkerBridge (or an equivalent queue around launcher.handleRequest) so the fix is obvious from the warning as well.

packages/server-mcp/src/browser/server/mcp-worker-bridge.ts
martin-fleck-at: Nice catch! Added a feature-detect so that Firefox/Safari fall back to buffering the body into an ArrayBuffer (still transferable) with a warn-once.

martin-fleck-at:
@tortmayr Great, thank you! I pushed a follow-up commit to address your concerns

tortmayr (approved):
LGTM!

PR #149, opened 2026-08-25
GLSP-1727: Keep applying computed bounds when an entry cannot be applied

martin-fleck-at:
#### What it does

A single unresolvable element in a ComputedBoundsAction aborted the whole batch: the client had already measured every element, but one stale id threw and the server dropped the remaining bounds, alignments and routes along with the model update the client was waiting for.

- resolve elements via index.find / index.findByClass so an unknown id yields undefined instead of throwing
- report a route with fewer than two points as not applicable rather than as an error, a client may have nothing to report for an unmeasured edge yet
- log a skipped entry and still submit the model
- log skipped routes at debug, not warn, an edge the client has not finished routing is expected on the initial layout pass
- split the apply step into applyElementBounds, applyAlignments and applyRoutes so adopters can override one kind on its own
- leave applyRoutingPoints strict, an unknown id in ChangeRoutingPointsOperationHandler is a real error

applyRoute now returns GEdge | undefined; noted in the changelog under potentially breaking changes.

Relates to eclipse-glsp/glsp#1727

#### How to test

1. In the workflow example, add an element to the diagram root from a client-side feedback command without giving it the feedbackFeature.
2. Open the diagram, so the server issues a bounds request and the client reports the extra element back.
3. The server logs Could not retrieve element with id: … and the diagram never finishes loading.

#### Changelog

- [x] This PR introduces a breaking change (if yes, provide more details below for the changelog and the migration guide)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@
 
 ### Potentially Breaking Changes
 
+- [layout] Keep applying computed bounds when an individual entry cannot be applied [#149](https://github.com/eclipse-glsp/glsp-server-node/pull/149)
+    - `applyRoute` now returns `GEdge | undefined` instead of `GEdge`
+    - `applyElementAndBounds`, `applyAlignment` and `applyRoute` no longer throw for an element the index cannot resolve, they report it as not applied. `applyRoutingPoints` stays strict.
+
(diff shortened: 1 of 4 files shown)


tortmayr (approved):
LGTM! Thanks 👍🏼

PR #150, opened 2026-08-25
GLSP-1729: Expose the bound address and accepted connections

### LINKED ISSUE #1729: Expose the port the socket server launcher bound to
When embedding a GLSP server it is common to bind port `0` and let the OS assign a free port, then tell the client where to connect. There is currently no programmatic way to find out which port was chosen.

The launcher resolves the port and then only writes it to stdout as part of the startup message, [here](https://github.com/eclipse-glsp/glsp-server-node/blob/79035df6eee608a14015ef71cadd9923b787a62e/packages/server/src/node/launch/socket-server-launcher.ts#L45-L61). [`start()`](https://github.com/eclipse-glsp/glsp-server-node/blob/79035df6eee608a14015ef71cadd9923b787a62e/packages/server/src/common/launch/glsp-server-launcher.ts#L42-L48) resolves to `void`, and [`netServer`](https://github.com/eclipse-glsp/glsp-server-node/blob/79035df6eee608a14015ef71cadd9923b787a62e/packages/server/src/node/launch/socket-server-launcher.ts#L27) is `protected`, so the only supported route is parsing the startup line out of stdout.

That works for a spawned process but is awkward in practice: it means racing against interleaved log output and stream buffering, and it is unavailable when the launcher is hosted in-process rather than spawned. Adopters therefore tend to subclass and reach into the protected field instead.

### What we would like

An accessor or event on `SocketServerLauncher` that reports the resolved address once the server is listening, so an embedder can pass the port on to its client without scraping output.

### Alternatives considered

- Parsing the startup message from stdout. Works today, but fragile, and not an option for in-process hosting.
- Subclassing to access `netServer` directly. What adopters do now, but it depends on an internal.

### Minor aside

On a socket error the launcher [calls `shutdown()` without logging anything](https://github.com/eclipse-glsp/glsp-server-node/blob/79035df6eee608a14015ef71cadd9923b787a62e/packages/server/src/node/launch/socket-server-launcher.ts#L62-L66), unlike the two address-resolution failures just above it, which do log before shutting down. The error itself is not lost, since the promise returned from `run()` rejects with it, but the server's own log gives no indication of why it stopped. Small consistency fix if someone is in the file anyway.

martin-fleck-at:
#### What it does

Launching with port 0 lets the operating system pick a free port, but the launcher kept both that port and the server socket to itself. An embedder that needs to hand the port to a client, or to observe accepted connections, had to parse the startup message off stdout or reach for the protected server field.

- add `listening`, a deferred that resolves with the bound address once the socket is up and rejects if it never gets there
- add `port`, a snapshot of the port currently bound, undefined while not listening
- add an `onConnection` event to the socket and websocket launchers, fired for every accepted socket before the JSON-RPC connection is built on it, so a listener can still see the client handshake
- arm a fresh deferred per run, so a settled promise is never handed to the next launch
- split the socket event handling into onListening, onError, failToListen and acceptConnection so adopters can override one step
- report a socket error through the logger before shutting down, it was previously discarded

Exposing this surfaced that a restarted launcher could not be shut down: the launchers registered their cleanup in the constructor, but DisposableCollection.dispose pops every entry it holds, so the first shutdown emptied the collection and the second reported success while the server socket stayed bound and the client sessions stayed alive.

- add a registerDisposables hook to GLSPServerLauncher, called once per launch, and move the netServer, websocket server and server-instance cleanup into it
- disposing a launcher that was never started is now a no-op instead of closing an undefined server

Fixes eclipse-glsp/glsp#1729

#### How to test

Everything is working as before but now you can add listeners and get the port returned properly.

#### Changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@
 
 ### Changes
 
+- [launch] Expose the address a `SocketServerLauncher` bound to via `listening` and `port`, so an embedder no longer has to parse the startup message to learn an OS-assigned port [#150](https://github.com/eclipse-glsp/glsp-server-node/pull/150)
+- [launch] Add an `onConnection` event to `SocketServerLauncher` and `WebSocketServerLauncher`, so an embedder can observe accepted connections without reaching for the protected server field [#150](https://github.com/eclipse-glsp/glsp-server-node/pull/150)
+- [launch] Release the server socket again when a restarted launcher is shut down [#150](https://github.com/eclipse-glsp/glsp-server-node/pull/150)
+    - launchers now register what `shutdown` has to release in the new `registerDisposables` hook, called once per launch, rather than in their constructor
+
 ### Potentially Breaking Changes
 
 - [layout] Keep applying computed bounds when an individual entry cannot be applied [#149](https://github.com/eclipse-glsp/glsp-server-node/pull/149)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,16 +4,19 @@
 
 ### Changes
 
-- [launch] Expose the address a `SocketServerLauncher` bound to via `listening` and `port`, so an embedder no longer has to parse the startup message to learn an OS-assigned port [#150](https://github.com/eclipse-glsp/glsp-server-node/pull/150)
+- [launch] Expose the address a launcher bound to via `listening` and `port` on `JsonRpcGLSPServerLauncher`, so an embedder no longer has to parse the startup message to learn an OS-assigned port [#150](https://github.com/eclipse-glsp/glsp-server-node/pull/150)
 - [launch] Add an `onConnection` event to `SocketServerLauncher` and `WebSocketServerLauncher`, so an embedder can observe accepted connections without reaching for the protected server field [#150](https://github.com/eclipse-glsp/glsp-server-node/pull/150)
+    - the web socket event carries the upgrade request next to the socket, which is the only place its headers and query are still available
 - [launch] Release the server socket again when a restarted launcher is shut down [#150](https://github.com/eclipse-glsp/glsp-server-node/pull/150)
-    - launchers now register what `shutdown` has to release in the new `registerDisposables` hook, called once per launch, rather than in their constructor
+    - `WebSocketServerLauncher` now also closes the HTTP server it mounts on, which `ws` leaves listening because it did not create it
 
 ### Potentially Breaking Changes
 
 - [layout] Keep applying computed bounds when an individual entry cannot be applied [#149](https://github.com/eclipse-glsp/glsp-server-node/pull/149)
     - `applyRoute` now returns `GEdge | undefined` instead of `GEdge`
     - `applyElementAndBounds`, `applyAlignment` and `applyRoute` no longer throw for an element the index cannot resolve, they report it as not applied. `applyRoutingPoints` stays strict.
+- [launch] Launchers register what `shutdown` has to release in the new `GLSPServerLauncher.registerDisposables` hook, called once per launch, rather than in their constructor [#150](https://github.com/eclipse-glsp/glsp-server-node/pull/150)
+    - A custom launcher that pushes into `toDispose` from its constructor keeps compiling but loses that cleanup after the first `shutdown`, because `dispose` empties the collection. Move those registrations into an override of `registerDisposables`.
 
 ## [v2.7.0 - 01/06/2026](https://github.com/eclipse-glsp/glsp-server-node/releases/tag/v2.7.0)
 
--- a/packages/server/src/node/launch/socket-server-launcher.ts
+++ b/packages/server/src/node/launch/socket-server-launcher.ts
@@ -1,5 +1,5 @@
 /********************************************************************************
- * Copyright (c) 2022-2024 STMicroelectronics and others.
+ * Copyright (c) 2022-2026 STMicroelectronics and others.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0 which is available at
@@ -13,7 +13,7 @@
  *
  * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
  ********************************************************************************/
-import { Deferred, Disposable, Emitter, Event } from '@eclipse-glsp/protocol';
+import { Disposable, Emitter } from '@eclipse-glsp/protocol';
 import { inject, injectable } from 'inversify';
 import * as net from 'net';
 import * as jsonrpc from 'vscode-jsonrpc/node';
@@ -24,9 +24,7 @@ import { Logger } from '../../common/utils/logger';
 export class SocketServerLauncher extends JsonRpcGLSPServerLauncher<net.TcpSocketConnectOpts> {
     @inject(Logger) protected override logger: Logger;
 
-    protected netServer: net.Server;
-
-    protected listeningAddress = new Deferred<net.AddressInfo>();
+    protected netServer?: net.Server;
 
     protected onConnectionEmitter = new Emitter<net.Socket>();
 
@@ -36,49 +34,32 @@ export class SocketServerLauncher extends JsonRpcGLSPServerLauncher<net.TcpSocke
      * A listener must not consume the socket. Subscribe before {@link start} to observe the first connection; the
      * subscription outlives an individual launch and is disposed by the caller.
      */
-    get onConnection(): Event<net.Socket> {
-        return this.onConnectionEmitter.event;
-    }
+    readonly onConnection = this.onConnectionEmitter.event;
 
-    /**
-     * Resolves with the address the server bound to once it is listening, and rejects if it never gets there.
-     *
-     * Launching with port `0` lets the operating system pick a free port, so the bound address is not necessarily the
-     * requested one. Settles once per launch and keeps that value, so use {@link port} for the current state.
-     */
-    get listening(): Promise<net.AddressInfo> {
-        return this.listeningAddress.promise;
-    }
-
-    /**
-     * The port the server is currently listening on, or `undefined` if it is not.
-     *
-     * Await {@link listening} rather than polling this while the server is still starting up.
-     */
-    get port(): number | undefined {
-        const address = this.netServer?.address();
-        return address && typeof address !== 'string' ? address.port : undefined;
+    protected override getServerSocket(): net.Server | undefined {
+        return this.netServer;
     }
 
     protected override registerDisposables(): void {
         super.registerDisposables();
         this.toDispose.push(
             Disposable.create(() => {
-                this.netServer.close();
+                this.netServer?.close();
             })
         );
     }
 
     protected run(opts: net.TcpSocketConnectOpts): Promise<void> {
         this.resetListening();
-        this.netServer = net.createServer(socket => this.acceptConnection(socket));
+        const netServer = net.createServer(socket => this.acceptConnection(socket));
+        this.netServer = netServer;
 
-        this.netServer.listen(opts.port, opts.host);
-        this.netServer.on('listening', () => this.onListening());
-        this.netServer.on('error', error => this.onError(error));
+        netServer.listen(opts.port, opts.host);
+        netServer.on('listening', () => this.handleListening(netServer));
+        netServer.on('error', error => this.handleError(error));
         return new Promise((resolve, reject) => {
-            this.netServer.on('close', () => resolve(undefined));
-            this.netServer.on('error', error => reject(error));
+            netServer.on('close', () => resolve(undefined));
+            netServer.on('error', error => reject(error));
         });
     }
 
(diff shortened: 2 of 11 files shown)


packages/server/src/node/launch/socket-server-launcher.ts
```diff
@@ -36,36 +70,74 @@ export class SocketServerLauncher extends JsonRpcGLSPServerLauncher<net.TcpSocke
...
+     * @param socket The accepted socket.
+     */
+    protected acceptConnection(socket: net.Socket): void {
+        this.onConnectionEmitter.fire(socket);
+        this.createServerInstance(this.createConnection(socket));
+    }
+
+    /** Arms {@link listening} for a launch, so a settled promise is never handed to the next one. */
+    protected resetListening(): void {
+        this.listeningAddress = new Deferred<net.AddressInfo>();
```
tortmayr: Swapping in a new `Deferred` here orphans the one from the field initializer, so a promise grabbed before `start` never settles. `const p = launcher.listening; launcher.start(opts); await p;` hangs (checked locally). `Deferred` exposes `state`, so only replacing it when it is no longer `'unresolved'` makes both orderings work.

packages/server/src/node/launch/websocket-server-launcher.ts
```diff
@@ -50,8 +50,20 @@ export class WebSocketServerLauncher extends JsonRpcGLSPServerLauncher<WebSocket
...
+     */
+    get onConnection(): Event<WebSocket> {
+        return this.onConnectionEmitter.event;
+    }
+
+    protected override registerDisposables(): void {
+        super.registerDisposables();
         this.toDispose.push(
             Disposable.create(() => {
                 this.server.close();
```
tortmayr: `this.server` is only assigned after the `await this.resolveOptions(...)` in `run`, and `resolveOptions` does a full listen/close cycle in `getFreePort`. A `shutdown()` in that window throws a TypeError here, `DisposableCollection.dispose` catches it and stops, so the server-instance cleanup pushed before this one is silently skipped. `this.server?.close()`.

tortmayr: `ws` only closes an HTTP server it created itself, and here one is handed in via `{ server }`, so this leaves the listener from `createHttpServer` bound and every restart adds another. The changelog claim "Release the server socket again when a restarted launcher is shut down" doesn't hold for the websocket launcher.

CHANGELOG.md
```diff
@@ -4,6 +4,11 @@
 
 ### Changes
 
+- [launch] Expose the address a `SocketServerLauncher` bound to via `listening` and `port`, so an embedder no longer has to parse the startup message to learn an OS-assigned port [#150](https://github.com/eclipse-glsp/glsp-server-node/pull/150)
+- [launch] Add an `onConnection` event to `SocketServerLauncher` and `WebSocketServerLauncher`, so an embedder can observe accepted connections without reaching for the protected server field [#150](https://github.com/eclipse-glsp/glsp-server-node/pull/150)
+- [launch] Release the server socket again when a restarted launcher is shut down [#150](https://github.com/eclipse-glsp/glsp-server-node/pull/150)
+    - launchers now register what `shutdown` has to release in the new `registerDisposables` hook, called once per launch, rather than in their constructor
```
tortmayr: This is an adopter-facing contract change: a custom launcher that pushes into `toDispose` from its constructor keeps compiling but silently loses its cleanup after a restart. Belongs under Potentially Breaking Changes with a one-line migration hint.

packages/server/src/node/launch/socket-server-launcher.ts
```diff
@@ -36,36 +70,74 @@ export class SocketServerLauncher extends JsonRpcGLSPServerLauncher<net.TcpSocke
...
+        // Nobody is obliged to consume `listening`, so pre-handle the rejection of a launch whose failure is not awaited.
+        this.listeningAddress.promise.catch(() => undefined);
+    }
+
+    /** Reports the server as ready and resolves {@link listening}, or fails if the bound address cannot be used. */
+    protected onListening(): void {
```
tortmayr: `onConnection` is an `Event` on this class, but `onListening` and `onError` are plain handlers, so `onX` means two different things here. `handleListening` / `handleError` would keep it unambiguous.

packages/server/src/node/launch/socket-server-launcher.ts
```diff
@@ -36,36 +70,74 @@ export class SocketServerLauncher extends JsonRpcGLSPServerLauncher<net.TcpSocke
...
+    protected acceptConnection(socket: net.Socket): void {
+        this.onConnectionEmitter.fire(socket);
+        this.createServerInstance(this.createConnection(socket));
+    }
+
+    /** Arms {@link listening} for a launch, so a settled promise is never handed to the next one. */
+    protected resetListening(): void {
+        this.listeningAddress = new Deferred<net.AddressInfo>();
+        // Nobody is obliged to consume `listening`, so pre-handle the rejection of a launch whose failure is not awaited.
+        this.listeningAddress.promise.catch(() => undefined);
```
tortmayr: The `Deferred` constructor already attaches `promise.then(res => ..., rej => ...)`, so the rejection is never unhandled (verified with a standalone script). This line and its comment can go.

packages/server/src/node/launch/socket-server-launcher.ts
```diff
@@ -26,8 +26,42 @@ export class SocketServerLauncher extends JsonRpcGLSPServerLauncher<net.TcpSocke
 
     protected netServer: net.Server;
 
-    constructor() {
-        super();
+    protected listeningAddress = new Deferred<net.AddressInfo>();
+
+    protected onConnectionEmitter = new Emitter<net.Socket>();
```
tortmayr, line 29: `AbstractMcpServerLauncher` declares these as `protected onXEmitter = new Emitter(); readonly onX = this.onXEmitter.event;` ([link](https://github.com/eclipse-glsp/glsp-server-node/blob/60e76e34fd805be12d7bac7fade19c0039757c44/packages/server-mcp/src/common/server/abstract-mcp-server-launcher.ts#L169-L172)). Same shape here keeps the launchers aligned. The emitter is also never disposed, so listeners survive `dispose()`.

packages/server/src/node/launch/websocket-server-launcher.ts
```diff
@@ -50,8 +50,20 @@ export class WebSocketServerLauncher extends JsonRpcGLSPServerLauncher<WebSocket
...
-        super();
+    protected onConnectionEmitter = new Emitter<WebSocket>();
+
+    /**
+     * Fires for every web socket the server accepts, before the JSON-RPC connection is built on it.
+     *
+     * A listener must not consume the socket. Subscribe before {@link start} to observe the first connection; the
+     * subscription outlives an individual launch and is disposed by the caller.
+     */
+    get onConnection(): Event<WebSocket> {
```
tortmayr: This launcher gets `onConnection` but not `listening` / `port`, even though it is the one that picks its own free port ([getFreePort](https://github.com/eclipse-glsp/glsp-server-node/blob/60e76e34fd805be12d7bac7fade19c0039757c44/packages/server/src/node/launch/websocket-server-launcher.ts#L113-L125)). Same discovery problem, so either add them here or lift them to `JsonRpcGLSPServerLauncher`.

packages/server/src/node/launch/websocket-server-launcher.ts
```diff
@@ -66,17 +78,24 @@ export class WebSocketServerLauncher extends JsonRpcGLSPServerLauncher<WebSocket
...
             this.server.on('error', error => reject(error));
         });
     }
 
+    /**
+     * Takes up a web socket the server accepted. Observers are notified before the socket is read from.
+     *
+     * @param socket The accepted web socket.
+     */
+    protected acceptConnection(socket: WebSocket): void {
```
tortmayr: The `req` (`IncomingMessage`) is dropped. For websockets the handshake is already over when `connection` fires, so `req` is the only way an observer can see it, which is exactly what the PR description promises.

packages/server/src/node/launch/socket-server-launcher.ts
```diff
@@ -26,8 +26,42 @@ export class SocketServerLauncher extends JsonRpcGLSPServerLauncher<net.TcpSocke
...
+        return this.listeningAddress.promise;
+    }
+
+    /**
+     * The port the server is currently listening on, or `undefined` if it is not.
+     *
+     * Await {@link listening} rather than polling this while the server is still starting up.
+     */
+    get port(): number | undefined {
+        const address = this.netServer?.address();
```
tortmayr: `netServer` is declared non-optional but needs `?.` here because it really is undefined before the first launch. Declaring it `netServer?: net.Server` matches reality and forces the same guard in `registerDisposables`.

packages/server/src/node/launch/socket-server-launcher.ts
tortmayr, line 1: `pnpm headers:check` fails on the three touched sources (this file, `websocket-server-launcher.ts` and `common/launch/glsp-server-launcher.ts`), the copyright end year has to be 2026.

tortmayr (changes requested):

packages/server/src/node/launch/socket-server-launcher.ts
martin-fleck-at: Fixed. `resetListening` now only swaps in a new `Deferred` when the current one has already settled, so a promise grabbed before `start` is the one that launch resolves. Added a test for that ordering. Without the fix it hangs and times out, which matches what you saw.

packages/server/src/node/launch/websocket-server-launcher.ts
martin-fleck-at: Fixed, it is `this.server?.close()` now. I also made `run` bail out when a shutdown arrives while the options are still resolving. Otherwise the guard only avoids the crash, and `run` would still go on to bind a socket that nothing closes.

martin-fleck-at: Correct, and the changelog line was wrong. The launcher now keeps the HTTP server in its own field and closes it on shutdown. Covered by a new websocket spec, one test for the restart case and one for a shutdown that lands mid startup. Changelog updated to say what actually happens.

Also switched to the `WebSocketServer` export. `ws` only ships the `Server` alias from its CommonJS entry, so the old import was unresolvable from ESM, which is what blocked a websocket spec.

CHANGELOG.md
martin-fleck-at: Moved to Potentially Breaking Changes, with a migration hint pointing adopters at an override of `registerDisposables`.

packages/server/src/node/launch/socket-server-launcher.ts
martin-fleck-at: Renamed to `handleListening` and `handleError`. `handleListening` now also takes the server it was called for, which avoids a non-null assertion once the field became optional.

packages/server/src/node/launch/socket-server-launcher.ts
martin-fleck-at: Removed, both the line and the comment. The constructor does attach the handlers, so there was nothing to guard against.

packages/server/src/node/launch/socket-server-launcher.ts
martin-fleck-at: Switched to the same `readonly onX = this.onXEmitter.event` shape as the MCP launcher.

I left the emitter undisposed though. `shutdown` goes through `stop` to `dispose`, so disposing it there would also kill the subscription on a restart, which is documented and covered by the "keeps notifying observers across a restart" test. There is no end of life hook separate from shutdown at the moment. Happy to add one if you want the emitter cleaned up.

packages/server/src/node/launch/websocket-server-launcher.ts
martin-fleck-at: Lifted both to `JsonRpcGLSPServerLauncher`. `http.Server` extends `net.Server`, so `net.AddressInfo` already covers both and no new common type was needed. A subclass just says which server to read from via `getServerSocket`. That one has a default returning undefined rather than being abstract, so a custom launcher extending the base still compiles.

packages/server/src/node/launch/websocket-server-launcher.ts
martin-fleck-at: `acceptConnection` takes the request now and the event payload carries it, so observers can actually see it. The websocket payload is `{ socket, request }`, while the socket launcher stays a plain socket since node only hands over the socket there. The new spec asserts the query string arrives on the request.

packages/server/src/node/launch/socket-server-launcher.ts
martin-fleck-at: Done, it is `netServer?: net.Server` now, and the disposable guards with `?.`.

packages/server/src/node/launch/socket-server-launcher.ts
martin-fleck-at: Fixed, all three are on 2026 and `checkHeaders` is clean. Worth noting that CI does not run the header check for this repo, only build, lint, test and e2e, which is why it stayed green.

martin-fleck-at:
Fixed and rebased on main.

tortmayr (approved):

PR #152, opened 2026-08-27
Correct MCP tool result contracts and registration

martin-fleck-at:
A review of the MCP tool handlers in `@eclipse-glsp/server-mcp` surfaced a set of
result-contract and input-schema defects. This fixes them and closes the class of
bug at the type level.

## Result contracts

The SDK validates `CallToolResult.structuredContent` against the declared
`outputSchema` and, on a mismatch, replaces the entire result with an error
result. `create-edges` with `dryRun: true` omitted the required
`dispatchedCommands`, so the dry run never returned its verdicts — the LLM saw
`Output validation error: Invalid structured content for tool create-edges`.

To stop that recurring, the tool handler bases take an optional output type
parameter `O`, and both `outputSchema` and the `success()` payload are typed
against it. That closes the loop in both directions: emitting a payload that
does not satisfy the schema fails to compile (removing the fix now fails `tsc`
with `Property 'dispatchedCommands' is missing`), and declaring a schema whose
shape differs from `O` fails too (`Property 'commandsRedone' is missing`).

## Behaviour

- `save-model` checked `isDirty` before considering `fileUri`, so a "save as" to
  a new location on a clean model reported success and wrote nothing.
- `modify-nodes` accepted any `GShapeElement`, but core's
  `GModelChangeBoundsOperationHandler` resolves via `findByClass(elementId, GNode)`
  and returns early for anything else — a compartment or port silently no-opped
  while the tool reported success.
- `undo` / `redo` dispatched N times without re-checking the stack and reported
  the requested count as the count actually applied. Their inputs also allowed
  non-integers, which then failed the `.int()` output schema.
- `validate-diagram` and `set-view` alias-resolved ids without an existence
  check; `index.getAll` drops unknown ids silently, so a hallucinated id came
  back as an empty marker list that reads as "diagram is clean".
- `modify-edges` counted entries that requested no change as modified, and a
  failure on one entry dropped a second entry's success for the same edge.
- `set-selection` documented an empty array as the way to clear a selection,
  which the shared `.min(1)` schema fragment rejected.
- The `create-*` tools inferred the new element by filtering the id diff on
  `type === elementTypeId`, reporting a false "creation likely failed" for
  adopters whose handler builds a different concrete type.

## Registration

`canRegister()` was answering two questions at once: whether a dependency is
bound for the diagram type, and whether the connected GLSP client supports an
action. Only the first is knowable when the MCP catalog is built, and the
catalog consulted neither — so `layout` was advertised on servers with no
`LayoutEngine` and failed at call time.

The hooks are now split. `isSupportedByDiagramType()` is evaluated once per
diagram type during harvest, against a container with that type's modules
loaded, and gates catalog registration. `canRegister()` is unchanged and still
gates the per-session registry, since the harvest container has no client and
would read every client capability as absent — which is why the export
resources must keep using it.

Verified against the real workflow modules: all 16 handlers resolve at harvest,
and `layout` is registered with `ElkLayoutModule` and dropped without it.

## Coverage

Handler specs called `createResult` directly and cast the result, so nothing
exercised the declared output schema — which is why the `create-edges` defect
survived. Specs now round-trip `structuredContent` through the handler's own
`outputSchema`, plus tests for the empty-array selection form and the
registration gate.


packages/server-mcp/src/common/tools/handlers/diagram-model-mcp-tool-handler.ts
```diff
@@ -57,8 +58,8 @@ export class DiagramModelMcpToolHandler extends AbstractMcpDiagramToolHandler<Di
     protected createResult({ sessionId }: DiagramModelInput): McpToolResult {
         const root = this.modelState.root;
         const structured = this.serializer.serializeStructured(root);
-        const count = Array.isArray(structured.elements) ? structured.elements.length : 0;
-        return this.success(this.summarizeModel(root, count), { sessionId, ...structured });
+        const elements = (Array.isArray(structured.elements) ? structured.elements : []) as DiagramModelOutput['elements'];
```
tortmayr, line 61: This isn't type-neutral: the SDK only validates `structuredContent`, it doesn't strip or re-emit the parsed object, so extra top-level keys an adopter's `serializeStructured` returns used to reach the client and now get dropped. `query-elements` keeps the spread for the same serializer contract ([link](https://github.com/eclipse-glsp/glsp-server-node/blob/9a56831afeac75862afe0e7526729123c345baa8/packages/server-mcp/src/common/tools/handlers/query-elements-mcp-tool-handler.ts#L106-L109)), so the two sibling handlers now disagree. Either spread here too, or drop the spread there.

packages/server-mcp/src/common/server/mcp-diagram-handler-dispatcher.ts
```diff
@@ -123,15 +130,56 @@ export class DefaultMcpDiagramHandlerDispatcher implements McpDiagramHandlerDisp
...
+     * Fails open: a constructor that cannot be resolved here stays in the catalog.
+     */
+    protected filterSupported<C extends interfaces.Newable<DiagramTypeSupportAware>>(
+        container: Container,
+        constructors: C[],
+        diagramType: string,
+        defaultHook: DiagramTypeSupportAware['isSupportedByDiagramType']
+    ): C[] {
+        return constructors.filter(constructor => {
+            if (constructor.prototype.isSupportedByDiagramType === defaultHook) {
```
tortmayr: A constructor whose prototype has no `isSupportedByDiagramType` at all (duck-typed handler, not extending the base) fails this identity check, gets resolved through DI, throws `TypeError: ... is not a function`, and lands in the fail-open `catch` with a misleading "Could not probe" line. A `typeof constructor.prototype.isSupportedByDiagramType !== 'function'` short-circuit next to the identity check keeps it out of the probe entirely.

packages/server-mcp/src/common/tools/handlers/modify-nodes-mcp-tool-handler.ts
```diff
@@ -88,53 +96,74 @@ export class ModifyNodesMcpToolHandler extends OperationMcpDiagramToolHandler<Mo
             throw new McpToolError(`modify-nodes does not accept edges — got: ${wrongType.join(', ')}. Use modify-edges for edges.`);
         }
 
-        // Reject any other non-shape kinds (labels, compartments, ports, custom kinds) — they
-        // reach the index too and would silently produce no-op or undefined-bounds operations.
-        const nonShape = elements.filter(([, element]) => !(element instanceof GShapeElement)).map(([change]) => `'${change.elementId}'`);
-        if (nonShape.length) {
+        // Core's `GModelChangeBoundsOperationHandler` only applies bounds to a `GNode`
+        // (`findByClass`). Label-only edits stay open to every element kind.
+        const unmovable = elements
```
tortmayr, line 110: Narrowing the guard to bounds-only also widens what the tool accepts for label edits: `GGraph`/`GModelRoot` is not a `GShapeElement`, so it used to be rejected by the old blanket check ([link](https://github.com/eclipse-glsp/glsp-server-node/blob/9a56831afeac75862afe0e7526729123c345baa8/packages/server-mcp/src/common/tools/handlers/modify-nodes-mcp-tool-handler.ts#L93-L98)) and now passes for a `text`-only entry. On a diagram with a free-floating top-level `GLabel`, `{ elementId: '<root>', text: 'x' }` renames that label and reports the root as modified.

packages/server-mcp/src/common/tools/handlers/test/expect-structured-content.ts
```diff
@@ -0,0 +1,29 @@
...
+import { expect } from 'vitest';
+import { ZodObject, ZodRawShape } from 'zod/v4';
+import { McpToolResult } from '../../../server/mcp-handler-shared';
+
+/**
+ * Asserts that a tool result satisfies the handler's declared `outputSchema`. The MCP SDK runs
+ * this check on every `tools/call` and replaces the result with an error result when it fails,
+ * which a spec calling `createResult` directly does not exercise.
+ */
+export function expectValidStructuredContent(schema: ZodObject<ZodRawShape>, result: McpToolResult): void {
```
tortmayr: Taking the schema as a parameter means a spec can pass the wrong schema and still go green, which is the failure mode this helper exists to prevent. Take the handler and read `handler.outputSchema` instead; it's public and the matrix test already relies on that link ([link](https://github.com/eclipse-glsp/glsp-server-node/blob/9a56831afeac75862afe0e7526729123c345baa8/packages/server-mcp/src/common/server/mcp-tool-handler.spec.ts#L173-L174)).

packages/server-mcp/src/common/server/mcp-tool-handler.ts
```diff
@@ -95,7 +95,8 @@ export abstract class BaseMcpToolHandler {
      * Optional dual-emit schema. When set, pass the matching `structured` payload to
      * {@link success} so the framework forwards it as `structuredContent` alongside the
      * human-readable text. The MCP spec says clients SHOULD validate `structuredContent`
-     * against the declared schema, so the two MUST stay in sync.
+     * against the declared schema, so the two MUST stay in sync. Bind the `O` type parameter to
+     * `z.infer<typeof MyOutputSchema>` to have the compiler enforce that.
      */
     readonly outputSchema?: ZodObject<ZodRawShape>;
```
tortmayr: The new `O` parameter isn't tied to this field, so a subclass can declare `O = FooOutput` while assigning `outputSchema = BarOutputSchema` and still compile. Typing it as something like `ZodObject<ZodRawShape> & ZodType<O>` would actually close the loop the PR description claims (`.shape` access in `toRegistrationConfig` still works with the intersection).

packages/server-mcp/src/common/tools/handlers/set-selection-mcp-tool-handler.ts
```diff
@@ -18,14 +18,18 @@ import { ActionDispatcher, SelectAction, SelectAllAction } from '@eclipse-glsp/s
 import { inject, injectable } from 'inversify';
 import * as z from 'zod/v4';
 import { McpToolResult } from '../../server/mcp-handler-shared';
-import { McpDiagramScopedInputSchema, elementIds } from '../../server/mcp-input-schemas';
+import { McpDiagramScopedInputSchema } from '../../server/mcp-input-schemas';
 import { AbstractMcpDiagramToolHandler } from '../../server/mcp-tool-handler';
 
 export const SetSelectionInputSchema = McpDiagramScopedInputSchema.extend({
-    selectedElementIds: elementIds
+    selectedElementIds: z
```
tortmayr: This diverges from the shared fragment convention documented right above `elementIds` ([link](https://github.com/eclipse-glsp/glsp-server-node/blob/9a56831afeac75862afe0e7526729123c345baa8/packages/server-mcp/src/common/server/mcp-input-schemas.ts#L21-L22)) without saying why, and the reason isn't obvious (Zod can't relax an existing `.min(1)`). Either add a shared empty-allowing fragment next to `elementIds`, or leave a one-liner here pointing at the constraint.

packages/server-mcp/src/common/tools/handlers/set-selection-mcp-tool-handler.spec.ts
```diff
@@ -97,14 +103,24 @@ describe('SetSelectionMcpToolHandler', () => {
...
+        });
 
+        expectValidStructuredContent(SetSelectionOutputSchema, result);
         expect(dispatched).toHaveLength(1);
         const action = dispatched[0] as SelectAction;
         expect(action.selectedElementsIDs).toEqual(['a', 'b']);
         expect(action.deselectedElementsIDs).toEqual(['c']);
     });
 
+    it('accepts the documented empty-array form for clearing the selection', () => {
```
tortmayr: The test name says "for clearing the selection" but this only parses the schema, it never runs the handler or asserts that `SelectAllAction.create(false)` gets dispatched. Either rename it to reflect that it's a schema test, or drive it through `callCreateResult` like the case above.

CHANGELOG.md
```diff
@@ -9,6 +9,19 @@
...
+- [mcp] Reject unknown element ids in `validate-diagram` and `set-view`, which previously dropped them and returned an empty, clean-looking result [#152](https://github.com/eclipse-glsp/glsp-server-node/pull/152)
+- [mcp] Keep tools out of the MCP catalog when no diagram type supports them, so `layout` is no longer advertised without a bound `LayoutEngine` [#152](https://github.com/eclipse-glsp/glsp-server-node/pull/152)
+    - the new `isSupportedByDiagramType()` hook on the diagram tool and resource bases covers statically bound dependencies; `canRegister()` keeps gating capabilities of the connected GLSP client
+- [mcp] Write the `validate-diagram` dedup separator as a `\u001f` escape, so the source file is no longer classified as binary by git [#152](https://github.com/eclipse-glsp/glsp-server-node/pull/152)
+- [mcp] Align tool schemas and descriptions with what the tools actually accept and apply [#152](https://github.com/eclipse-glsp/glsp-server-node/pull/152)
+    - `set-selection` accepts the documented empty-array form for clearing the selection, and `undo` / `redo` require integer counts
```
tortmayr, line 21: This one is source-file hygiene rather than user-facing behaviour, adopters see no difference. I'd drop it from the changelog and keep it in the commit message.

tortmayr (changes requested):

martin-fleck-at:
Thanks Tobias, all eight are addressed and pushed.

Two of them were more right than I realised. The duck typed handler probe and the outputSchema gap were both already visible in this PR's own tests, I just hadn't noticed. The schema binding is now closed in both directions, I checked by deliberately declaring the wrong schema and confirming it no longer compiles.

The NUL bytes turned out to predate this PR, they came in with 2.7.0, but I fixed them here anyway since the file was open. Dropped that line from the changelog as you suggested.

One thing I did differently: for set-selection I added a shared elementIdsAllowingEmpty fragment next to elementIds instead of a comment at the call site, so there is no divergence left to explain.

tortmayr (approved):
Thanks for addressing all the issues.
LGMT! 👍🏼

